PR #88, opened 2020-02-19
Dev class assignments

varunasingh:
Method and Model signatures for Views, Entities, Daos, Presenters and Tests.


core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzAssignmentDetailPresenter.kt
```diff
@@ -0,0 +1,65 @@
...
+    private var clazzAssignmentDao = repository.clazzAssignmentDao
+    private var clazzAssignment : ClazzAssignmentWithMetrics ? = null
+    private lateinit var factory: DataSource.Factory<Int, PersonWithAssignmentMetrics>
+
+    override fun onCreate(savedState: Map<String, String?>?) {
+        super.onCreate(savedState)
+        if(arguments.containsKey(ARG_CLAZZ_ASSIGNMENT_UID)){
+            val clazzAssignmentUid = arguments[ARG_CLAZZ_ASSIGNMENT_UID]!!.toLong()
+            GlobalScope.launch {
+                clazzAssignment = clazzAssignmentDao.findWithMetricByUidAsync(clazzAssignmentUid)
```
mikedawson: Better to use ?: 0. We should avoid !!

mikedawson:
This is obsolete and replaced with #159

PR #107, opened 2020-03-15
Dev fix assignment test

varunasingh:


mikedawson:
This is replaced with #159

PR #124, opened 2020-05-06
Dev crud school

varunasingh:


app-android/src/main/java/com/ustadmobile/port/android/view/MainActivity.kt
```diff
@@ -78,11 +81,7 @@ class MainActivity : AppCompatActivity(), UstadListViewActivityWithFab, NavContr
 
         val currentFrag =
                 findNavController(R.id.activity_main_navhost_fragment).currentDestination?.id?:0
-        if(BOTTOM_NAV_DEST.contains(currentFrag)){
-            menu.findItem(R.id.menu_main_settings).setVisible(true)
-        }else{
-            menu.findItem(R.id.menu_main_settings).setVisible(false)
-        }
+        menu.findItem(R.id.menu_main_settings).isVisible = BOTTOM_NAV_DEST.contains(currentFrag)
```
mikedawson, line 84: :+1: I like this - it is more concise

app-android/src/main/java/com/ustadmobile/port/android/view/SchoolEditFragment.kt
```diff
@@ -0,0 +1,219 @@
...
+        clazzRecyclerView?.adapter = clazzRecyclerAdapter
+        clazzRecyclerView?.layoutManager = LinearLayoutManager(requireContext())
+
+
+        mPresenter = SchoolEditPresenter(requireContext(), arguments.toStringMap(), this,
+                this, UstadMobileSystemImpl.instance,
+                UmAccountManager.getActiveDatabase(requireContext()),
+                UmAccountManager.getRepositoryForActiveAccount(requireContext()),
+                UmAccountManager.activeAccountLiveData)
+        //mPresenter?.onCreate(savedInstanceState.toNullableStringMap())
```
mikedawson, line 106: Can this commented out line be removed? Was that from the template?

app-android/src/main/res/layout/fragment_school_overview.xml
```diff
@@ -0,0 +1,183 @@
...
+            name="editButtonMode"
+            type="com.ustadmobile.core.view.EditButtonMode"/>
+
+
+    </data>
+
+    <androidx.core.widget.NestedScrollView
+        android:id="@+id/fragment_school_detail_overview_nestedscrollview"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
```
mikedawson, line 30: layout_height here should be match_parent - e.g. the nested scroll view should match the height of the parent, and then the scrolling should be within it (eg as it's child items layout_height is wrap_content)

app-android/src/main/res/layout/item_clazz_simple.xml
```diff
@@ -0,0 +1,68 @@
...
+
+        <androidx.appcompat.widget.AppCompatImageView
+            android:id="@+id/item_clazz_simple_secondary_menu_imageview"
+            android:layout_width="wrap_content" android:layout_height="wrap_content"
+            android:layout_marginTop="@dimen/list_twoline1_margin_vertical"
+            android:layout_marginEnd="@dimen/list_twoline1_margin_horizontal"
+            app:layout_constraintEnd_toEndOf="parent" app:layout_constraintTop_toTopOf="parent"
+            app:srcCompat="@drawable/ic_delete_black_24dp"
+            android:contentDescription="@string/delete"
+            android:visibility="invisible"
```
mikedawson, line 63: Why is this invisible?

app-android/src/main/res/layout/item_school_list_item.xml
```diff
@@ -0,0 +1,122 @@
...
+        <variable
+            name="school"
+            type="com.ustadmobile.lib.db.entities.SchoolWithMemberCountAndLocation" />
+        <variable
+            name="presenter"
+            type="com.ustadmobile.core.controller.SchoolListPresenter" />
+
+    </data>
+
+    <androidx.cardview.widget.CardView
```
mikedawson, line 16: I think we should not use CardView in lists so that we keep things consistent.

core/locale/main/values/strings_ui.xml
```diff
@@ -384,6 +385,7 @@
     <string name="first_names">First names</string>
     <string name="basic_details">Basic details</string>
     <string name="home_address">Home address</string>
+    <string name="address">Address</string>
```
mikedawson, line 388: Perhaps we should remove home_address so that we just use address on person detail?

core/src/commonMain/kotlin/com/ustadmobile/core/controller/SchoolDetailOverviewPresenter.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    override suspend fun onLoadEntityFromDb(db: UmAppDatabase): SchoolWithHolidayCalendar? {
+        val entityUid = arguments[ARG_ENTITY_UID]?.toLong() ?: 0L
+
+         val schoolWithHolidayCalendar = withTimeoutOrNull(2000) {
+             db.schoolDao.findByUidWithHolidayCalendarAsync(entityUid)
+         } ?: SchoolWithHolidayCalendar()
+
+        if(entityUid != 0L) {
```
mikedawson, line 42: I think we don't need this if statement. entityUid for the detail presenter should never be 0. This presenter is loading from the database.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/SchoolDetailPresenter.kt
```diff
@@ -1,60 +1,49 @@
...
-        val args = mapOf(ARG_SCHOOL_DETAIL_SCHOOL_UID to currentSchool.schoolUid.toString())
-        impl.go(SchoolEditView.VIEW_NAME, args, context)
+    override suspend fun onLoadEntityFromDb(db: UmAppDatabase): School? {
+        val entityUid = arguments[ARG_ENTITY_UID]?.toLong() ?: 0L
+
+        val school =  withTimeoutOrNull(2000) {
+            db.schoolDao.findByUidAsync(entityUid)
+        } ?: School()
+
+        view.setTitle(school.schoolName?:"")
```
mikedawson, line 40: I think to keep things consistent it would be better to use an abstract var with set/get methods on the view instead of having a method. E.g. then we would have view.title = school?.schoolName ?: ""

lib-database-entities/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/School.kt
```diff
@@ -28,6 +28,12 @@ open class School() {
     //Active
     var schoolActive: Boolean = false
 
+    var schoolPhoneNumber : Long = 0L
```
mikedawson, line 31: This should be a string to allow for + symbol, areas codes brackets etc. e.g. +1 (212) 123 4567

app-android/src/main/res/layout/fragment_school_overview.xml
```diff
@@ -0,0 +1,183 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mikedawson, line 1: Can you please add the vertical margins between fields? We have a vertical form item margin constant in the dimens.xml

core/src/commonMain/kotlin/com/ustadmobile/core/view/SchoolDetailView.kt
```diff
@@ -2,22 +2,17 @@ package com.ustadmobile.core.view
...
 import com.ustadmobile.lib.db.entities.School
 
-/**
- * Core View. Screen is for SchoolDetail's View
- */
-interface SchoolDetailView : UstadView {
 
-    //Updates school name on view and sets edit gear icon to visible and feature visibility
-    fun setSchool(school: School)
+interface SchoolDetailView: UstadDetailView<School> {
```
mikedawson, line 6: I think the FAB should be managed by the Overview or List fragment.

core/src/commonMain/kotlin/com/ustadmobile/core/view/SchoolDetailView.kt
```diff
@@ -2,22 +2,17 @@ package com.ustadmobile.core.view
...
 
     fun setSettingsVisible(visible: Boolean)
 
-    fun setUpTabs(tabs: List<String>)
+    fun setTitle(title: String)
```
mikedawson, line 10: Let's change this to a var instead of a function so it is consistent with other views.

core/src/commonMain/kotlin/com/ustadmobile/staging/core/view/SchoolEditView.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    fun setPicture(picturePath: String)
+
+    //Add any other view bits here..
+
+    companion object {
+        // This defines the view name that is an argument value in the go() in impl.
+        const val VIEW_NAME = "SchoolDetail"
+        const val ARG_SCHOOL_DETAIL_SCHOOL_UID = "ArgSchoolDetailSchoolUid"
+        const val ARG_SCHOOL_NEW = "ArgSchoolNew"
```
mikedawson, line 26: Why does the edit presenter need this argument for whether or not a school is new? If it is new, then no uid is provided.

core/src/commonMain/kotlin/com/ustadmobile/staging/core/view/SchoolEditView.kt
```diff
@@ -0,0 +1,30 @@
...
+    fun setSchool(school: School)
+
+    fun setPicture(picturePath: String)
+
+    //Add any other view bits here..
+
+    companion object {
+        // This defines the view name that is an argument value in the go() in impl.
+        const val VIEW_NAME = "SchoolDetail"
+        const val ARG_SCHOOL_DETAIL_SCHOOL_UID = "ArgSchoolDetailSchoolUid"
```
mikedawson, line 25: We should be using the UstadView.ARG_ENTITY_UID here - I think this argument constant shouldn't be needed.

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/UmAppDatabase.kt
```diff
@@ -38,11 +38,17 @@ import kotlin.jvm.Volatile
     //TODO: DO NOT REMOVE THIS COMMENT!
     //#DOORDB_TRACKER_ENTITIES
 
-], version = 36)
+], version = 37)
```
mikedawson, line 41: Let's not increase the database version here. There are other fields that need migrated. I think we are going to have to a diff of lib-database-entities before we merge dev-crud. That would show us what fields have been added/removed.

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/ClazzDao.kt
```diff
@@ -66,6 +67,26 @@ abstract class ClazzDao : BaseDao<Clazz> {
...
+
+    @Query("SELECT * FROM Clazz WHERE clazzSchoolUid = :schoolUid " +
+            "AND CAST(isClazzActive AS INTEGER) = 1 ")
+    abstract fun findAllClazzesBySchoolLive(schoolUid: Long)
+            : DataSource.Factory<Int,Clazz>
+
+    @Query("UPDATE Clazz SET isClazzActive = :active WHERE clazzUid = :clazzUid ")
+    abstract suspend fun updateActiveByClazzUid(clazzUid: Long, active : Boolean)
+
+    override suspend fun deactivateByUids(uidList: List<Long>) {
```
mikedawson, line 83: If someone deactivates a class from the school page, I think this should just break the connection between it and the school, not deactivate the whole class. We also wouldn't want to do that without some kind of double confirmation (e.g. dialog). I think this should set clazzSchoolUid = 0 instead of setting active to false.

PR #127, opened 2020-05-18
Data model for Clazz work

varunasingh:
Just the models


mikedawson (commented):
My comment is re. the comment entities. I think we should have one comment entity. Comments are not specific to a submission. All private comments (e.g. those that are to/from the teacher) should be shown on the assignment. That isn't separate.

I think the entity should just be 'Comment'. I think we could (and probably should) reuse the comments in the future on other items (e.g. comments on content, discussion boards, etc).

varunasingh:
https://youtu.be/WkCsG-0CD6Y

mikedawson:
This pull request is now obsolete and replaced with #159

PR #131, opened 2020-05-27
Content Entry migration

kileha3:
Converted content entry list, edit and details screen to CURD template
Demo: https://www.youtube.com/watch?v=-JEgMjbPlEk


app-android/src/main/res/navigation/mobile_navigation.xml
```diff
@@ -116,6 +118,36 @@
...
+        android:id="@+id/home_content_dest"
+        android:name="com.ustadmobile.port.android.view.ContentEntryListTabsFragment"
+        android:label="@string/content">
+        <argument
+            android:name="parentUid"
+            android:defaultValue="-4103245208651563007L"
+            app:argType="string" />
+    </fragment>
+
+    <dialog
```
mikedawson, line 131: Are we using this as a navigation destination?

core/locale/main/values/strings_ui.xml
```diff
@@ -253,7 +266,7 @@
     <string name="error_message_update_document">Failed to update document</string>
     <string name="content_update_massage">Updated successfully</string>
     <string name="content_folder_image_message">Support for adding an image is coming soon</string>
-    <string name="content_supported_files">Supported files: <![CDATA[<b>Epub, TinCan (XAPI) Zip</b>]]></string>
+    <string name="content_supported_files">Supported files: <![CDATA[<b>Video, Audio, EPUB, Xapi Zip,H5P</b>]]></string>
```
mikedawson, line 269: Can we keep this as plain text? The HTML in CDATA can confuse translators.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntry2DetailPresenter.kt
```diff
@@ -0,0 +1,122 @@
...
+        systemImpl.go(ContentEntryEdit2View.VIEW_NAME, mapOf(ARG_ENTITY_UID to entity?.contentEntryUid.toString()), context)
+    }
+
+    fun handleOnClickOpenDownloadButton(){
+        val canOpen = !isDownloadEnabled || downloadJobItemLiveData?.getValue()?.djiStatus == JobStatus.COMPLETE
+        if (canOpen) {
+            val loginFirst = systemImpl.getAppConfigString(AppConfig.KEY_LOGIN_REQUIRED_FOR_CONTENT_OPEN,
+                    "false", context)!!.toBoolean()
+
+            if (loginFirst) {
```
mikedawson, line 80: This needs to check if the user is already logged in. The current implementation will always take the user to the login screen.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntry2DetailPresenter.kt
```diff
@@ -0,0 +1,122 @@
...
+        } else if (isDownloadEnabled) {
+            view.showDownloadOptionsDialog(arguments)
+        }
+    }
+
+    private fun onDownloadJobItemChanged(downloadJobItem: DownloadJobItem?) {
+        if(downloadJobItem != null){}
+    }
+
+    private fun goToSelectedContentEntry() {
```
mikedawson, line 94: Let's call this openContentEntry. It's not really the selectedContentEntry - there is no selection taking place on this screen.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -0,0 +1,165 @@
...
+                       view.setUpStorageOptions(result)
+                   })
+                }
+            }
+        })
+    }
+
+    override suspend fun onLoadEntityFromDb(db: UmAppDatabase): ContentEntryWithLanguage? {
+        val entityUid = arguments[ARG_ENTITY_UID]?.toLong() ?: 0
+        return withTimeoutOrNull(2000) {
```
mikedawson, line 86: As per comment on the Fragment for this, let's change this to set the leaf property according to the argument

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -0,0 +1,165 @@
...
+
+            if(canCreate){
+                if(entity.leaf){
+                    entity.licenseName = systemImpl.getString(LicenceOptions.values()
+                            .single { it.optionVal == entity.licenseType }.messageId, context)
+                }
+
+                if(entity.contentEntryUid == 0L) {
+                    entity.contentEntryUid = repo.contentEntryDao.insertAsync(entity)
+                    val contentEntryJoin = ContentEntryParentChildJoin()
```
mikedawson, line 123: In Kotlin this could be written as:

val contentEntryJoin = ContentEntryParentChildJoin().apply {
    cepcjChildContentEntryUid = entity.contentEntryUid
    cepcjParentContentEntryUid = parentEntryUid
}

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryList2View.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.ustadmobile.lib.db.entities.ContentEntryWithParentChildJoinAndStatusAndMostRecentContainer
+
+interface ContentEntryList2View: UstadListView<ContentEntry, ContentEntryWithParentChildJoinAndStatusAndMostRecentContainer> {
+
+    enum class ContentEntryListViewMode(val viewMode: Int) {
+        NORMAL(1),
+        PICKER(2)
+    }
+
+    fun checkAndGetContentType(type: String?): String
```
mikedawson, line 13: To be removed

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryList2View.kt
```diff
@@ -0,0 +1,30 @@
+package com.ustadmobile.core.view
+
+import com.ustadmobile.lib.db.entities.ContentEntry
+import com.ustadmobile.lib.db.entities.ContentEntryWithParentChildJoinAndStatusAndMostRecentContainer
+
+interface ContentEntryList2View: UstadListView<ContentEntry, ContentEntryWithParentChildJoinAndStatusAndMostRecentContainer> {
+
+    enum class ContentEntryListViewMode(val viewMode: Int) {
```
mikedawson, line 8: Why does this have it's own enum instead of using the ones provided by UstadView?

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryList2View.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    fun checkAndGetContentDrawable(type: String?): Int
+
+    companion object {
+
+        const val ARG_DOWNLOADED_CONTENT = "downloaded"
+
+        const val ARG_CONTENT_FILTER = "filter"
+
+        const val ARG_LIBRARIES_CONTENT = "libraries"
```
mikedawson, line 23: This should be ARG_FILTER_PARENT_UID

sharedse/src/commonJvmMain/kotlin/com/ustadmobile/port/sharedse/contentformats/ContentTypeUtil.kt
```diff
@@ -80,6 +83,8 @@ suspend fun importContentEntryFromFile(file: File, db: UmAppDatabase, dbRepo: Um
  */
 object ContentTypeUtil {
 
+
```
mikedawson, line 86: Can we remove object ContentTypeUtil ?

core/src/commonMain/kotlin/com/ustadmobile/core/impl/ShowErrorUmCallback.kt
```diff
@@ -1,17 +1,17 @@
 package com.ustadmobile.core.impl
 
-import com.ustadmobile.core.view.UstadViewWithSnackBar
+import com.ustadmobile.core.view.UstadBaseFeedbackMessageView
```
mikedawson, line 3: This class is unused and being deleted on dev-crud-nav

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryAddOptionsFragment.kt
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Fragment class responsible for content creation selection, you can create content by one of the following
+ * CONTENT_CREATE_FOLDER = Create new content category
+ * CONTENT_IMPORT_FILE = Create content from file (epub, h5p e.t.c)
+ * CONTENT_CREATE_CONTENT = create content from out content editor
+ * CONTENT_IMPORT_LINK = Create content from a link
+ */
+
+class ContentEntryAddOptionsFragment : UstadBottomSheetFragment(), ContentEntryAddOptionsView, View.OnClickListener {
```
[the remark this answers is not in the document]
kileha3: I think there is, this is the same as other base fragments we have which are subclasses of  Fragment itself. i.e DialogFragment. FYI, We'll be using BottomSheetDialogFragment on ContentEditor and e.tc. That's why

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntry2DetailFragment.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    override fun handleFileSize(fileSize: Long): String {
+        return UMFileUtil.formatFileSize(fileSize)
+    }
+
+    override fun handleOnClickOpenDownloadButton() {
+        mPresenter?.handleOnClickOpenDownloadButton()
+    }
+
+    override fun setAvailableTranslations(result: List<ContentEntryRelatedEntryJoinWithLanguage>?) {
```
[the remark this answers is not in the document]
mikedawson: Small change request: can you change currentLiveData to currentTranslationLiveData . CurrentLiveData on it's own in this context might make people think it is the entity itself.

mikedawson:
ContentEntryAddOptionsBottomSheetFragment has a memory leak. If you click on "Add content", then go back, and then go to another item on the bottom navigation, a memory leak notification will appear.

It says library leak, but I think this is the AddOptionsBottomSheetFragment. It's probably because of setting the onClickListener.

I would recommend keeping the view as a variable, and then calling setOnClickListener(null) to remove the leak.

mikedawson:
ContentEntry2DetailFragment has a memory leak. I have found that when using a recyclerview inside a fragment, you must explicitly set the adapter to null in onDestroyView. 

RecyclerViewAdapters must imlpement onDetachedFromRecyclerView and explicitly null-out any references they hold that might be pointing to the fragment, view etc (e.g. the presenter variable).

mikedawson:
In ContentEntryEdit2Presenter: I think this code is repeating itself:

```
when {
                    entity.title == null && view.selectedFileUri != null -> view.titleErrorEnabled = true
                    entity.title != null && view.selectedFileUri == null && entity.leaf -> view.fileImportErrorVisible = true
                    else -> {
                        view.titleErrorEnabled = true
                        if(entity.leaf){
                            view.fileImportErrorVisible = true
                        }
                    }
                }
```

mikedawson:
Can you please add JVM test for ContentEntryList and ContentEntryDetail using the templates?

PR #132, opened 2020-06-15
Dev crud report

SamihMustafa:


mikedawson:
It seems like the legend is showing without a key (top right square with the color but no text) [see video](https://devserver3.ustadmobile.com/devsvrstatic/jobs/ustadMobile-dev-nav-report/builds/9/htmlreports/ADB_20Screen_20Record/com.ustadmobile.port.android.view.ReportDetailFragmentTest/givenReportExists_whenLaunched_thenShouldShowReport/emulator-5554.mp4) .

app-android/src/main/res/layout/fragment_report_detail.xml
```diff
@@ -0,0 +1,35 @@
...
+            type="boolean" />
+
+        <variable
+            name="fieldsEnabled"
+            type="boolean" />
+
+    </data>
+
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
mikedawson, line 23: There is no need for the constraintlayout here. The recyclerview should be the root view. Here the recyclerview width and height is set to match parent, so the recyclerview will recycle views as intended. When recyclerviews are nested, and height is set to wrap_content, the recycling stops working and this triggers loading everything at once.

app-android/src/main/res/layout/item_clazz_log_attendance_list.xml
```diff
@@ -67,7 +67,7 @@
 
         <!-- Line 1-->
         <TextView
-            android:id="@+id/line1_text"
+            android:id="@+id/report_title"
```
mikedawson, line 70: Is there a reason that this file (from attendance view) was modified? Seems like just the rename of a textview id.

core/locale/main/values/strings_ui.xml
```diff
@@ -308,11 +309,19 @@
     <string name="xapi_options_what">What</string>
     <string name="xapi_options_when">When</string>
     <string name="xapi_options_where">Where</string>
+    <string name="xapi_options_content">Content</string>
```
mikedawson, line 312: We need to avoid string id prefixes so that we easily (and by default) use the same string in multiple parts of the app. We don't want to duplicate the same string (and create additional translation work). We should change 'xapi_options_what' to just 'what'. I think we will already have the word content from somewhere else in the app, so we don't need to add it.

core/locale/main/values/strings_ui.xml
```diff
@@ -308,11 +309,19 @@
...
+
+    <string name="report">Report</string>
+    <string name="verb">Verb</string>
+    <string name="id_verbentity_urlid">ID: %s</string>
+    <string name="verb_filter">verb filter</string>
```
mikedawson, line 317: Could we have a string substitution instead? E.g. 
```
<string name="x_filter">%1$s filter</string>
```
Then that string can be used to create Person filter, content filter, etc. That would reduce the number of words that are part of the translation.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ReportEditPresenter.kt
```diff
@@ -0,0 +1,264 @@
...
+            view.titleErrorText = null
+        }
+
+
+        GlobalScope.launch(doorMainDispatcher()) {
+
+            if (entity.reportUid != 0L) {
+                repo.reportDao.updateAsync(entity)
+
+                repo.reportFilterDao.insertListAsync(personOneToManyJoinEditHelper.entitiesToInsert)
```
mikedawson, line 196: You could make reportFilterDao implement OneToManyDao. Then you could simply call :
```
personOneToManyJoinEditHelper.commitToDatabase(repo.reportFilterDao) {
   reportFk = entity.reportUid
}
```
As you are calling the same method on a list of things, why not do something like:
```
listOf(personOneToManyJoinEditHelper, verbDisplaynOneToManyJoinEditHelper, contentOneToManyJoinEditHelper).forEach {
   it.commitToDatabase(repo.reportFilterDao)
}

```

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ReportEditPresenter.kt
```diff
@@ -0,0 +1,264 @@
...
+
+                repo.reportDao.updateAsync(entity)
+
+                withContext(doorMainDispatcher()) {
+                    view.finishWithResult(listOf(entity))
+                }
+
+            } else {
+
+                val personUidList = personOneToManyJoinEditHelper.liveList.getValue()?.map { it.entityUid }
```
mikedawson, line 216: The commitToDatabase function above can also work. Once you have the entity saved to the database, the code should be the same.  The commitToDatabaseFunction can be used to set the foreign key.

core/src/jvmTest/kotlin/com/ustadmobile/core/controller/ReportDetailPresenterTest.kt
```diff
@@ -0,0 +1,191 @@
...
+        val presenterArgs = mapOf(ARG_ENTITY_UID to testEntity.reportUid.toString())
+        val presenter = ReportDetailPresenter(context,
+                presenterArgs, mockView, mockLifecycleOwner,
+                systemImplRule.systemImpl, clientDbRule.db, clientDbRule.repo,
+                clientDbRule.accountLiveData)
+
+        presenter.onCreate(null)
+
+        //wait for the entity value to be set
+        mockView.captureLastEntityValue(120000)
```
mikedawson, line 183: Timeout should be shorter - e.g. 5 seconds is long enough for it to load one entity from the database. Otherwise if the test is failing, this could cause long delays.

mikedawson:
Can you add the vertical text view (axis label) to the report list screen test [video](https://devserver3.ustadmobile.com/devsvrstatic/jobs/ustadMobile-dev-nav-report/builds/11/htmlreports/ADB_20Screen_20Record/)

Issue #137, opened 2020-06-23
Migrate VideoContent screen to using a fragment


mikedawson:
Migrate VideoContent player to using a fragment

PR #143, opened 2020-06-23
Dev crud display content

SamihMustafa:


app-android/src/androidTest/java/com/ustadmobile/port/android/view/VideoContentFragmentTest.kt
```diff
@@ -0,0 +1,120 @@
...
+
+    @JvmField
+    @Rule
+    val dataBindingIdlingResourceRule = ScenarioIdlingResourceRule(DataBindingIdlingResource())
+
+    @JvmField
+    @Rule
+    val crudIdlingResourceRule = ScenarioIdlingResourceRule(CrudIdlingResource())
+
+
```
mikedawson, line 59: This needs the AdbScreenRecordRule for screen recording to run

app-android/src/androidTest/java/com/ustadmobile/port/android/view/VideoContentFragmentTest.kt
```diff
@@ -0,0 +1,120 @@
...
+    @Before
+    fun setup() {
+        val tmpDir = UmFileUtilSe.makeTempDir("testVideoPlayer",
+                "" + System.currentTimeMillis())
+
+        val videoFile = File(tmpDir, "video1.webm")
+
+        copyInputStreamToFile(
+                javaClass.getResourceAsStream("/com/ustadmobile/app/android/video.mp4")!!,
```
mikedawson, line 74: How big is that video.mp4 file?

app-android/src/androidTest/java/com/ustadmobile/port/android/view/VideoContentFragmentTest.kt
```diff
@@ -0,0 +1,120 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.io.File
+import java.lang.Thread.sleep
+
+
+@AdbScreenRecord("Video Content Screen Test")
+@ExperimentalStdlibApi
+class VideoContentFragmentTest {
```
mikedawson, line 41: This test needs to actually play the video and verify that the video plays (e.g. use onFragment to get access to the ExoPlayer instance and then get check that the status goes to playing.

app-android/src/main/java/com/ustadmobile/port/android/view/VideoContentFragment.kt
```diff
@@ -190,23 +196,25 @@ class VideoPlayerActivity : UstadBaseActivity(), VideoPlayerView {
...
+                subtitles?.visibility = View.GONE
 
                 player?.prepare(mediaSource, false, false)
             }
-
-
         }
     }
 
+    override fun showErrorWithAction(message: String, actionMessageId: Int) {
```
mikedawson, line 206: Couldn't the presenter just use showSnackBar directly?

app-android/src/main/java/com/ustadmobile/port/android/view/VideoContentFragment.kt
```diff
@@ -216,51 +224,52 @@ class VideoPlayerActivity : UstadBaseActivity(), VideoPlayerView {
...
+                if (container == null) {
                     showErrorWithAction(UstadMobileSystemImpl.instance.getString(MessageID.no_video_file_found, viewContext), 0)
                     return@launch
                 }
 
                 val containerManager = ContainerManager(container, appDatabase, repoAppDatabase)
 
                 val byteArrayDataSource = ByteArrayDataSource(
-                        UMIOUtils.readStreamToString(containerManager.getInputStream(containerManager.getEntry(subtitleData)!!)).toByteArray())
+                        UMIOUtils.readStreamToByteArray(containerManager.getInputStream(containerManager.getEntry(subtitleData)!!)))
```
mikedawson, line 237: There are **!!** here!! Let's remove those

sharedse/src/commonJvmMain/kotlin/com/ustadmobile/port/sharedse/contentformats/ContentTypeUtil.kt
```diff
@@ -66,9 +79,31 @@ suspend fun importContainerFromZippedFile(contentEntryUid: Long, mimeType: Strin
     return container
 }
 
+fun createContainerFromDirectory(directory: File, filemap: MutableMap<File, String>): Map<File, String> {
+    val sourceDirPath = Paths.get(directory.toURI())
+    try {
+        Files.walk(sourceDirPath).filter { path -> !Files.isDirectory(path) }
```
mikedawson, line 85: This won't be available on older versions of Android. We need to stick to java.io.File. 

All we are really supporting is adding one file as the only file of a container here. So we can do this straightforwardly.

sharedse/src/commonJvmMain/kotlin/com/ustadmobile/port/sharedse/contentformats/VideoTypePlugin.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.ustadmobile.lib.db.entities.ContentEntry
+import com.ustadmobile.lib.db.entities.Language
+import com.ustadmobile.port.sharedse.contentformats.ContentTypePlugin
+
+import java.io.File
+
+class VideoTypePlugin : VideoType(), ContentTypePlugin {
+
+    override fun getContentEntry(file: File): ContentEntryWithLanguage? {
+        return ContentEntryWithLanguage().apply {
```
mikedawson, line 14: This needs to check that the given file is actually a video file. For now let's do this by simply checking the file extension is a known video extension.

mikedawson:
I checked into the increase in method count and size here. I think it's nothing major. The increase in method count is because of more methods being referenced in the ExoPlayer package.

methodcounts

We can go back to using the vanillamultidex flavor which will work for this. The app size is still just 6.9MB.

app-android/src/androidTest/java/com/ustadmobile/port/android/view/VideoContentFragmentTest.kt
SamihMustafa: 359kb

mikedawson:
Can you please ensure that the naming of the component classes is as per our convention. We should have:

VideoContentFragment
VideoContentPresenter
VideoContentView

The presenter and the view are called VideoPlayer..

mikedawson:
On VideoContentView (currently VideoPlayerView) can you make sure that there is only set videoParams method? So that we follow the same convention as other views, can you please change this to being a data class with a variable? 

e.g. 

```
class VideoContentPresenter {
    data class VideoParams(val videoPath: String, val audioPath: String?, val srtLangList: MutableList<String>, val srtMap: MutableMap<String, String>)
}


interface VideoContentView {
   var videoParams: VideoContentPresenter.VideoParams
}
```

mikedawson:
Please rename importContainerFromZippedFile (ContentTypeUtil.kt) to importContainerFromFile. Calling 'importContainerFromZippedFile' with isZipped = false would not be intuitive for other developers.

mikedawson:
In VideoContentFragment we can't use the onResume provided by the fragment. This is actually tied to the lifecycle of the activity (e.g. it could be called when the fragment is inactive). You would need to observe viewLifecycleOwner and then run this in an onResume that comes from the viewlifecycle.

mikedawson:
In VideoContentFragment, in the GlobalScope.launch it would be better to just use Dispatchers.Main instead of runOnUiThread. 
e.g. 
```
withContext(Dispatchers.Main) {
player?.prepare(mergedSource, false, false)
}
```

mikedawson:
In VideoContentFragment onDestroyView needs to ensure that all player resources are released.

mikedawson:
In VideoTypePlugin please set the ContentEntry contentType flag to TYPE_VIDEO

mikedawson:
Implements issue #137

Issue #139, opened 2020-06-23
Migrate XapiPackageContent screen to Fragment


mikedawson:
Migrate the XapiPackageContent screen to using a Fragment

PR #144, opened 2020-06-25
Migrate XapiPackageContent screen to Fragment enhancement

kileha3:


app-android/src/main/java/com/ustadmobile/port/android/view/XapiPackageContentFragment.kt
```diff
@@ -0,0 +1,106 @@
...
+import com.ustadmobile.sharedse.network.NetworkManagerBle
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+
+class XapiPackageContentFragment : UstadBaseFragment(), XapiPackageContentView {
+
+    var networkManagerProvider: BleNetworkManagerProvider? = null
+
+    override fun onAttach(context: Context) {
```
mikedawson, line 28: Because networkManagerProvider is only accessible for testing purposes it would be good to use the internal modifier. Internal allows access only within the same module and the test code for that module.

app-android/src/main/java/com/ustadmobile/port/android/view/XapiPackageContentFragment.kt
```diff
@@ -0,0 +1,106 @@
...
+        super.onAttach(context)
+        if(context is BleNetworkManagerProvider){
```
mikedawson, line 30: A corresponding onDeatch function is needed otherwise there will be a memory leak.

app-android/src/main/java/com/ustadmobile/port/android/view/XapiPackageContentFragment.kt
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    override var contentTitle: String = ""
+        set(value) {
+            field = value
+            title = value
+        }
+
+    override var urlToLoad: String = ""
```
mikedawson, line 41: Rather than calling this urlToLoad, as it is a property, I think this can be simplified to just url

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntry2DetailPresenter.kt
```diff
@@ -94,7 +95,7 @@ class ContentEntry2DetailPresenter(context: Any,
     }
 
     private fun openContentEntry() {
-        GlobalScope.launch {
+        GlobalScope.launch(Dispatchers.Main) {
```
mikedawson, line 98: Well done for getting this one

core/src/jvmTest/kotlin/com/ustadmobile/core/controller/XapiPackageContentPresenterTest.kt
```diff
@@ -90,13 +89,12 @@ class XapiPackageContentPresenterTest {
...
         httpd!!.start()
-
-        mockXapiPackageContentView = mock(XapiPackageContentView::class.java)
+        mockedView = mock{}
 
         doAnswer { invocation ->
             Thread(invocation.getArgument<Any>(0) as Runnable).start()
             Any()
-        }.`when`<XapiPackageContentView>(mockXapiPackageContentView).runOnUiThread(any())
+        }.`when`<XapiPackageContentView>(mockedView).runOnUiThread(any())
```
mikedawson, line 97: The best way to do this in kotlin mockito is:
```
whenever(mockedView.runOnUiThread(any()).doAnswer { invocation ->
   Thread(invocation.getArgument<Any>(0) as Runnable).start()
}
```
You can add any if it needs that to finish the closure.

mikedawson:
Implements Issue #139

PR #145, opened 2020-06-26
Migrate EpubContent meet updated testing standards

kileha3:


mikedawson:
The navigation shelf does not seem respond to swiping from the side of the screen to appear or disappear. See [this video](https://www.youtube.com/watch?v=YdPDs_VfM90)

mikedawson:
Can you please change the default icon on the navigation drawer to use the same vector drawable that we have as the default for content entry leaves in ContentEntryList?

mikedawson:
Can you please use ViewPager2 to manage swiping from page to page? I think the old ViewPager is deprecated now. I'm not sure if we should be keeping a weak hashmap cache of fragments. It would be better to allow Android to manage the lifecycle to ensure we don't accidentally serve a destroyed fragment.

mikedawson:
Can we update the title when the user changes page?

PR #154, opened 2020-07-02
Login 

kileha3:


app-android/src/androidTest/java/com/ustadmobile/port/android/view/Login2FragmentTest.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    @JvmField
+    @Rule
+    val screenRecordRule = AdbScreenRecordRule()
+
+    @JvmField
+    @Rule
+    val crudIdlingResourceRule = ScenarioIdlingResourceRule(CrudIdlingResource())
+
+    private val context = getInstrumentation().targetContext
```
mikedawson: I think getInstrumentation is deprecated. We should use ApplicationProvider.getApplicationContext()

core/src/commonMain/kotlin/com/ustadmobile/core/controller/WorkspaceEnterLinkPresenter.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.Runnable
+import kotlinx.coroutines.launch
+
+class WorkspaceEnterLinkPresenter(context: Any, arguments: Map<String, String>, view: WorkspaceEnterLinkView,
+                                  val impl : UstadMobileSystemImpl = UstadMobileSystemImpl.instance) :
+        UstadBaseController<WorkspaceEnterLinkView>(context, arguments, view) {
+
+
+    fun goToNext(){
```
mikedawson: To follow the naming convention let's call this handleClickNext()

app-android/src/main/java/com/ustadmobile/port/android/view/GetStartedFragment.kt
```diff
@@ -0,0 +1,45 @@
+package com.ustadmobile.port.android.view
```
mikedawson, line 1: Can you please add the icons as per the prototype for this fragment? All icons appear the same in the video

core/src/commonMain/kotlin/com/ustadmobile/core/controller/WorkspaceEnterLinkPresenter.kt
```diff
@@ -0,0 +1,47 @@
...
+                        view.progressVisible = false
+                        view.validLink = false
+                    })
+                }
+            }
+        }
+    }
+
+    companion object{
+        const val VALID_WORKSPACE_DOMAIN = "ustadmobile.com/lms/"
```
mikedawson: We shouldn't require any specific URL - we should make a request to get an info object.

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -0,0 +1,100 @@
...
+            rootView = it.root
+            it.buttonEnabled = true
+            it.fieldsEnabled = true
+        }
+
+        mPresenter = Login2Presenter(requireContext(), UMAndroidUtil.bundleToMap(arguments),this,
+                personRepo = UmAccountManager.getRepositoryForActiveAccount(requireContext()).personDao)
+        mPresenter?.onCreate(savedInstanceState.toStringMap())
+
+        mBinding?.presenter = mPresenter
```
mikedawson, line 89: I would recommend to set mBinding.mPresenter before calling mPresenter.onCreate. Otherwise if the presenter calls back to do something, the presenter might not be set.

core/locale/main/values-ar/strings_ui.xml
```diff
@@ -6,7 +6,7 @@
     <string name="content_type_multimedia">وسائط متعددة (صوت و صورة)</string>
     <string name="multiple_choice">متعدد الخيارات</string>
     <string name="content_type_fill_the_blanks">إملأ الفراغات</string>
-    <string name="content_type_link">ربط</string>
+    <string name="link">ربط</string>
```
mikedawson: We must not change the translated xmls. These are generated by Weblate. Changing them might introduce a conflict. Can you please revert all non-english string xml files? Keep the change in English. Weblate will figure it out from there.

app-android/src/main/java/com/ustadmobile/port/android/view/WorkspaceEnterLinkFragment.kt
```diff
@@ -0,0 +1,111 @@
...
+        mPresenter?.onCreate(savedInstanceState.toStringMap())
+        mBinding?.organisationLink?.addTextChangedListener(object: TextWatcher {
+            override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) {}
+            override fun onTextChanged(url: CharSequence?, start: Int, before: Int, count: Int) {
+                inputCheckHandler.removeCallbacks(inputCheckerCallback)
+            }
+            override fun afterTextChanged(s: Editable?) {
+                val typedText = s.toString()
+                progressVisible = typedText.isNotEmpty()
+                val canValidate = PatternsCompat.WEB_URL.matcher(typedText).find()
```
mikedawson: Let's leave all validation in the common presenter.

app-android/src/main/java/com/ustadmobile/port/android/view/WorkspaceEnterLinkFragment.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    private val inputCheckHandler: Handler = Handler()
+
+    private var lastInputTime: Long = 0
+
+    @VisibleForTesting
+    private var isIdle = false
+
+    private val inputCheckerCallback = Runnable {
+        if (System.currentTimeMillis() > lastInputTime + inputCheckDelay) {
```
mikedawson: Why do we need to check the time again here? Changing the text would have canceled the callback. If there was something wrong with the calculation, checkLinkValidity would never be called.

app-android/src/main/java/com/ustadmobile/port/android/view/WorkspaceEnterLinkFragment.kt
```diff
@@ -0,0 +1,111 @@
...
+            override fun afterTextChanged(s: Editable?) {
+                val typedText = s.toString()
+                progressVisible = typedText.isNotEmpty()
+                val canValidate = PatternsCompat.WEB_URL.matcher(typedText).find()
+                        && typedText.contains(VALID_WORKSPACE_DOMAIN)
+                handleError(!canValidate)
+                if(!canValidate) return
+                handleIdling()
+                lastInputTime = System.currentTimeMillis()
+                inputCheckHandler.postDelayed(inputCheckerCallback, inputCheckDelay)
```
mikedawson, line 75: How about we show the progress spinner as soon as we get afterTextChanged? Idling Resources could just watch for the progress bar.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/Login2Presenter.kt
```diff
@@ -0,0 +1,112 @@
...
+                    AppConfig.KEY_API_URL, "http://localhost", context)?:""
+        }
+
+        view.createAccountVisible = impl.getAppConfigString(AppConfig.KEY_SHOW_CREATE_ACCOUNT,
+                "true", context)?.toBoolean() ?: true
+        view.connectAsGuestVisible = impl.getAppConfigString(AppConfig.KEY_SHOW_CONNECT_AS_GUEST,
+                "true", context)?.toBoolean() ?: true
+    }
+
+    fun handleLogin(username: String?, password:String?){
```
mikedawson, line 58: Please delegate this to UstadAccountManager instead of using the httpclient directly.

mikedawson (changes requested):

core/src/commonMain/kotlin/com/ustadmobile/core/controller/WorkspaceEnterLinkPresenter.kt
```diff
@@ -0,0 +1,47 @@
...
+class WorkspaceEnterLinkPresenter(context: Any, arguments: Map<String, String>, view: WorkspaceEnterLinkView,
+                                  val impl : UstadMobileSystemImpl = UstadMobileSystemImpl.instance) :
+        UstadBaseController<WorkspaceEnterLinkView>(context, arguments, view) {
+
+
+    fun goToNext(){
+        impl.go(Login2View.VIEW_NAME, mapOf(ARG_SERVER_URL to view.workspaceLink), context)
+    }
+
+    fun checkLinkValidity(){
```
mikedawson: Change to: handleCheckLinkText(href: String)

core/src/commonMain/kotlin/com/ustadmobile/core/controller/Login2Presenter.kt
```diff
@@ -0,0 +1,112 @@
...
+import io.ktor.client.request.get
+import io.ktor.client.request.parameter
+import io.ktor.client.statement.HttpStatement
+import io.ktor.http.HttpStatusCode
+import io.ktor.http.takeFrom
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.Runnable
+import kotlinx.coroutines.launch
+
+class Login2Presenter(context: Any, arguments: Map<String, String>, view: Login2View,
```
mikedawson, line 22: LoginPresenter should require the UstadAccountManager as a constructor parameter.

Issue #116, opened 2020-04-26
Migrate Class Assignments to using CRUD template / Fragment [0.5pts]


mikedawson:
The previously created assignment feature needs to be migrated to using the common CRUD parent classes / templates. It also needs to be migrated to use a fragment for compatibility with navigation controller.

PR #159, opened 2020-07-13
Dev crud classwork

varunasingh:


app-android/src/main/java/com/ustadmobile/port/android/view/binding/EditTextBindings.kt
```diff
@@ -86,3 +90,39 @@ fun TextInputEditText.setRunOnClickWhenFocused(runOnClickWhenFocused: Boolean) {
...
+fun EditText.setMinMax(min: String, max: Int){
+    filters =   arrayOf(InputFilterMinMax(Integer.valueOf(min), max))
+}
+
+
+class InputFilterMinMax(private val minimumValue: Int, private val maximumValue: Int) : InputFilter {
+    override fun filter(source: CharSequence, start: Int, end: Int, dest: Spanned, dstart: Int, dend: Int): CharSequence? {
+        try {
+            val input = (dest.subSequence(0, dstart).toString() + source + dest.subSequence(dend, dest.length)).toInt()
+            if (isInRange(minimumValue, maximumValue, input)) return null
```
mikedawson: Do we need isInRange? Couldn't this be written as:

```
if(input >= minimum && input <= maximumValue) return null

File: app-android/src/main/java/com/ustadmobile/port/android/view/binding/LinearLayoutBindings.kt [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+import androidx.databinding.BindingAdapter
+import com.toughra.ustadmobile.R
+import com.ustadmobile.lib.db.entities.ClazzWorkQuestionOption
+
+@BindingAdapter("questionOptions", "questionOptionsResponse")
+fun LinearLayout.setQuestionOptionsWithResponse(options: List<ClazzWorkQuestionOption>, response: Long) {
+
+    removeAllViews()
+    for (item in options) {
+        val option = TextView(context)
```
mikedawson: Just from a style point of view, I would use .apply here: e.g.

```
 val option = TextView(context).apply {
    layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
    text = item.clazzWorkQuestionOptionText
}
```

app-android/src/main/java/com/ustadmobile/port/android/view/binding/TextViewBindings.kt
```diff
@@ -204,3 +206,66 @@ fun TextView.setHtmlText(htmlText: String?) {
 fun TextView.setFileSize(fileSize: Long) {
     text = UMFileUtil.formatFileSize(fileSize)
 }
+
+@BindingAdapter(value=["clazzMemberWithClazzWorkAndProgress"])
+fun TextView.setClazzWorkMarking(clazzMemberWithClazzWorkAndProgress: ClazzMemberWithClazzWorkProgress){
+    val marked = context.getString(R.string.marked_cap)
```
mikedawson: Why not lookup these items as needed?

app-android/src/main/java/com/ustadmobile/port/android/view/binding/TextViewBindings.kt
```diff
@@ -204,3 +206,66 @@ fun TextView.setHtmlText(htmlText: String?) {
...
+                submitted
+            }
+        }else{
+            notSubmitted
+        }
+    }
+
+    if(clazzMemberWithClazzWorkAndProgress.clazzWorkHasContent &&
+            clazzMemberWithClazzWorkAndProgress.mProgress >= 0){
+        line2Bit2 = " - " + completed + " " + clazzMemberWithClazzWorkAndProgress.mProgress.toInt() +
```
mikedawson: I would recommend creating an extension function to get the status of a ClazzWorkSubmission: e.g.

```
fun ClazzWorkSubmission?.statusString(context: Context) = when {
    this == null -> context.getString(R.string.notSubmitted)
    this.clazzWorkSubmissionDateTimeMarked > 0 -> context.getString(R.string.marked)
    this.clazzWorkSubmissionDateTimeFinished > 0 -> context.getString(R.string.submitted)
    else -> context.getString(R.string.notSubmitted)

}
```

I would also use string templates. I think once you have that function - this whole binding adapter could become:
```
fun TextView.setClazzWorkMarking(clazzMemberWithClazzWorkAndProgress: ClazzMemberWithClazzWorkProgress){
   var line = "${clazzMemberWithClazzWorkAndProgress.mClazzWorkSubmission.statusString(context)}"
   if(clazzMemberWithClazzWorkAndProgress.clazzWorkHasContent && clazzMemberWithClazzWorkAndProgress.mProgress >= 0) {
        line += " ${context.getString(R.string.completed)} ${clazzMemberWithClazzWorkAndProgress.mProgress.toInt()}% ${context.getString(R.string.of_content)}"
    }
    text = line
}
```

app-android/src/main/java/com/ustadmobile/port/android/view/binding/TextViewBindings.kt
```diff
@@ -204,3 +206,66 @@ fun TextView.setHtmlText(htmlText: String?) {
...
+                "% " + ofContent
+    }
+    val line2 = line2Bit1 + line2Bit2
+
+    text = line2
+}
+
+@BindingAdapter(value=["selectedClazzWorkQuestionType"])
+fun TextView.setTypeText(clazzWorkQuestionType: Int){
```
mikedawson, line 226: This can be replaced with the MessageIDMap adapter

app-android/src/main/java/com/ustadmobile/port/android/view/binding/TextViewBindings.kt
```diff
@@ -204,3 +206,66 @@ fun TextView.setHtmlText(htmlText: String?) {
...
+fun TextView.setResponseTextFilled(responseText: String?){
+    if(responseText == null || responseText.isEmpty()){
+        text = context.getString(R.string.not_answered)
+    }else{
+        text = responseText
+    }
+}
+
+@BindingAdapter(value=["submissionButtonText"])
+fun Button.setSubmissionButtonText(more: Boolean){
```
mikedawson: Couldn't this be done in the layout xml itself using a conditional operator? E.g.

```
<Button 
   android:text="@{more ? @string/return_and_mark_next : @string/return_only}"
```

app-android/src/main/res/layout/item_clazzwork_progress_detail.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mikedawson, line 1: Can the textviews here be given ids?

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzWorkDetailOverviewPresenter.kt
```diff
@@ -0,0 +1,207 @@
...
+                clazzMember.clazzMemberRole == ClazzMember.ROLE_STUDENT)
+
+        //If Submission object doesnt exist, create it.
+        if(clazzWorkWithSubmission.clazzWorkSubmission == null){
+            clazzWorkWithSubmission.clazzWorkSubmission = ClazzWorkSubmission().apply {
+                clazzWorkSubmissionClazzWorkUid = clazzWorkWithSubmission.clazzWorkUid
+                clazzWorkSubmissionClazzMemberUid = clazzMember?.clazzMemberUid?:0L
+                clazzWorkSubmissionPersonUid = loggedInPersonUid
+                clazzWorkSubmissionInactive = false
+                clazzWorkSubmissionDateTimeStarted = UMCalendarUtil.getDateInMilliPlusDays(0)
```
mikedawson: Couldn't this be just:

```
getSystemTimeInMillis()
```

core/src/commonMain/kotlin/com/ustadmobile/core/controller/DefaultNewCommentItemListener.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun addComment(entityType: Int, entityUid: Long, comment: String,
+                            public: Boolean, to: Long, from: Long) {
+
+        val accountManager: UstadAccountManager by instance()
+
+        //val db: UmAppDatabase by di.activeRepoInstance()
+        val db: UmAppDatabase by on(accountManager.activeAccount).instance(tag = UmAppDatabase.TAG_DB)
+        val repo: UmAppDatabase by on(accountManager.activeAccount).instance(tag = UmAppDatabase.TAG_REPO)
+
+        val comment = Comments(entityType, entityId, fromPerson,
```
mikedawson: This already has a variable called comment in the function signature. Maybe call it commentObj?

core/src/commonMain/kotlin/com/ustadmobile/core/controller/DefaultNewCommentItemListener.kt
```diff
@@ -0,0 +1,30 @@
...
+                UMCalendarUtil.getDateInMilliPlusDays(0), comment, public)
```
mikedawson: If we're not adding anything to the date, can we replace getDateInMillisPlusDays with getSystemTimeInMillis()

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ClazzWithSchoolExt.kt
```diff
@@ -2,4 +2,16 @@ package com.ustadmobile.core.util.ext
 
 import com.ustadmobile.lib.db.entities.ClazzWithSchool
 
-fun ClazzWithSchool.effectiveTimeZone() = clazzTimeZone ?: school?.schoolTimeZone ?: "UTC"
+fun ClazzWithSchool.findClazzTimeZone(): String{
```
mikedawson: What is the difference between the objective of this and effectiveTimezone ? 

Also I think there is a logic error here:
if school != null and the timezone is null or blank, it would give UTC instead of looking at the class time zone.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzWorkDetailOverviewPresenter.kt
```diff
@@ -0,0 +1,207 @@
...
+                newOptionsAndResponse.add(everyResult)
+            }
+
+            val loggedInPersonUid = accountManager.activeAccount.personUid
+            val clazzMember: ClazzMember? = withTimeoutOrNull(2000){
+                db.clazzMemberDao.findByPersonUidAndClazzUid(loggedInPersonUid,
+                        entity?.clazzWorkClazzUid?:0L)
+            }
+
+            var submission = entity?.clazzWorkSubmission
```
mikedawson: I think it would be better style to use the elvis operator - e.g.

```
var submission = entity?.clazzWorkSubmission ?: ClazzWorkSubmission().apply {
   clazzWorkSubmissionClazzWorkUid = clazzWorkWithSubmission?.clazzWorkUid ?: 0L
   clazzWorkSubmissionClazzMemberUid = clazzMember?.clazzMemberUid ?: 0L
   clazzWorkSubmissionDateTimeFinished = getSystemTimeInMillis()
   clazzWorkSubmissionInactive = false
   clazzWorkSubmissionPersonUid = loggedInPersonUid
}
```

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzWorkDetailOverviewPresenter.kt
```diff
@@ -0,0 +1,207 @@
...
+        }?: ClazzWorkWithSubmission()
+
+        val clazzWithSchool = withTimeoutOrNull(2000){
+            db.clazzDao.getClazzWithSchool(clazzWorkWithSubmission.clazzWorkClazzUid)
+        }?: ClazzWithSchool()
+
+        view.timeZone = clazzWithSchool.findClazzTimeZone()
+
+        //Find Content and questions
+        val contentList = withTimeoutOrNull(2000) {
```
mikedawson, line 48: I think this setup is causing crashes. I am sometimes (maybe 1/4 times opening the page) seeing:

```
2020-07-24 00:31:34.198 22651-22651/com.toughra.ustadmobile E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.toughra.ustadmobile, PID: 22651
    java.lang.IllegalStateException: Can't access the Fragment View's LifecycleOwner when getView() is null i.e., before onCreateView() or after onDestroyView()
        at androidx.fragment.app.Fragment.getViewLifecycleOwner(Fragment.java:352)
        at com.ustadmobile.port.android.view.ClazzWorkDetailOverviewFragment.setClazzWorkContent(ClazzWorkDetailOverviewFragment.kt:354)
        at com.ustadmobile.core.controller.ClazzWorkDetailOverviewPresenter.onLoadEntityFromDb(ClazzWorkDetailOverviewPresenter.kt:54)
        at com.ustadmobile.core.controller.ClazzWorkDetailOverviewPresenter$onLoadEntityFromDb$1.invokeSuspend(Unknown Source:12)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)

```
I think the problem here is that this is really setting the result a datasource.factory. This should actually be done only once, and only from the repo (the datasource.factory updates automatically if something new is loaded). I think anything using a DataSource.Factory should be done as follows so we only set it once:

```
view.takeIf { it.clazzWorkContent == null }?.clazzWorkContent = repo.clazzWorkContentJoinDao.findAllContentByClazzWorkUidDF(
                clazzWorkWithSubmission.clazzWorkUid,
                loggedInPersonUid)
```

mikedawson:
I think there is an issue with private comments. Private comments from the teacher should be shown on the overview page whether or not the students can 'talk back' and send private comments to the teacher. The difference is that we would hide the submit field if this option is disabled.

mikedawson:
Looks like the tab visibility was set to view.hidden instead of view.gone when there is no tabs (e.g. student mode)

clazzwork-student-hilite

mikedawson:
If you open the assignment as a teacher, and switch tabs, the edit fab will disappear. 
The problem is in ClazzworkDetailFragment:

fabManagementEnabled should be set in onCreateView. It is set in onViewCreated, after the super call is made (so it's too late).

mikedawson:
If you go to an assignment as a teacher, then edit it, and then go back or click done, when you get back you get two blank tabs. I see a Handler is being used, so I'm not sure why that is.

mikedawson:
There are memory leaks being caused by the recycleradapters. Before making the adapter itself null, you need to detach it from the recyclerview: e.g.

```
mRecyclerView?.adapter = null
mAdapter = null
```

mikedawson (approved):

mikedawson:
Implements #116

PR #161, opened 2020-07-17
Person Registration

kileha3:


core/src/commonMain/kotlin/com/ustadmobile/core/controller/PersonAccountEditPresenter.kt
```diff
@@ -0,0 +1,110 @@
...
+import com.ustadmobile.lib.db.entities.Person
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlinx.serialization.json.Json
+import org.kodein.di.DI
+import org.kodein.di.instance
+
+
+class PersonAccountEditPresenter(context: Any,
```
mikedawson, line 23: I think for this to stick to the crud pattern we should add another object which holds the data that is going to be filled in. E.g. you could add:
```
class PersonWithAccount(val currentPassword: String, val newPassword: String, val passwordConfirmed: String)
```

The pattern we are following is that the presenter gives the object to the view. The object should generally have the properties that are going to be manipulated in the view.

core/src/commonMain/kotlin/com/ustadmobile/core/view/PersonAccountEditView.kt
```diff
@@ -0,0 +1,33 @@
+package com.ustadmobile.core.view
+
+import com.ustadmobile.lib.db.entities.Person
+
+interface PersonAccountEditView: UstadEditView<Person> {
+
+    var firstPassword: String?
```
mikedawson: Let's put these on an entity as per comment on presenter so this follows the pattern

core/src/commonMain/kotlin/com/ustadmobile/core/view/PersonAccountEditView.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    var secondPassword: String?
+
+    var secondPasswordFieldRequiredErrorVisible: Boolean
+
+    var firstPasswordFieldRequiredErrorVisible: Boolean
```
mikedawson: Couldn't we have passwordError: String? and confirmPasswordError: String?, then a null value means no error.

core/src/commonMain/kotlin/com/ustadmobile/core/view/PersonAccountEditView.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    fun showPasswordDoNotMatchError()
```
mikedawson: Couldn't this be done by the controller using:

```
view.errorMessage = impl.getString(MessageID.passwordsDontMatch)
```

core/src/commonMain/kotlin/com/ustadmobile/core/view/PersonAccountEditView.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    var usernameRequiredErrorVisible: Boolean
+
+    var errorMessage:String?
+
+    var fistPasswordFieldHint: String?
```
mikedawson: I think we should have three fixed password boxes : current password, new password, confirm new password. If the user is admin, we can hide the current password box.

core/src/jvmTest/kotlin/com/ustadmobile/core/controller/PersonAccountEditPresenterTest.kt
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @Test
+    fun givenPresenterCreated_whenNewPasswordIsNotFilledAndSaveClicked_thenShouldShowError(){
+        val person = createPerson(isAdmin = true)
+        val args = mapOf(UstadView.ARG_ENTITY_UID to person.personUid.toString())
+        val presenter = PersonAccountEditPresenter(context, args,mockView, di, mockLifecycleOwner)
+        presenter.onCreate(null)
+        presenter.handleClickSave(person)
+        argumentCaptor<Boolean>{
```
mikedawson: You don't need an argument captor for this kind of call. You could use just verify:

```
verify(mockView, timeout(defaultTimeout).atLeastOnce()).secondPasswordFieldRequiredErrorVisible = true
```
or use argument matchers
```
verify(mockView, timeout(defaultTimeout).atLeastOnce()).secondPasswordFieldRequiredErrorVisible = eq(true)
```

core/src/jvmTest/kotlin/com/ustadmobile/core/controller/PersonAccountEditPresenterTest.kt
```diff
@@ -0,0 +1,218 @@
...
+            on{firstPassword}.thenReturn("password")
+            on{secondPassword}.thenReturn("password")
+        }
+        val person = createPerson(isAdmin = true, withUsername = false)
+        val args = mapOf(UstadView.ARG_ENTITY_UID to person.personUid.toString(),
+                UstadView.ARG_SERVER_URL to serverUrl)
+        val presenter = PersonAccountEditPresenter(context, args,mockView, di, mockLifecycleOwner)
+        presenter.onCreate(null)
+        //wait for onLoadEntityFromDb to be called
+        sleep(1000)
```
mikedawson: We shouldn't use sleep as this would normally wait to long and sometimes not long enough. Even 1 second on every test will add up. You can use : captureLastEntityValue(timeout) ( which can be found in UstadSingleEntityView). 

Because the view is a mock, you can wait for the entity object to be set. E.g. this would also work:

```
verify(mockView, timeout(5000)).entity = any()
```

core/src/commonMain/kotlin/com/ustadmobile/core/controller/PersonEditPresenter.kt
```diff
@@ -2,38 +2,53 @@ package com.ustadmobile.core.controller
...
+
+    private val impl: UstadMobileSystemImpl by instance()
+
+    private  lateinit var nextDestination: String
+
     override val persistenceMode: PersistenceMode
         get() = PersistenceMode.DB
 
-    val clazzMemberJoinEditHelper = DefaultOneToManyJoinEditHelper<ClazzMemberWithClazz>(ClazzMemberWithClazz::clazzMemberUid,
+    private var workSpace: WorkSpace = WorkSpace().apply {
```
mikedawson: Why is the workspace object needed here? Couldn't we just have a simple boolean registrationMode as true/false ? It seems like we are using the Workspace class when all we need from this is the boolean.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/PersonEditPresenter.kt
```diff
@@ -55,6 +71,22 @@ class PersonEditPresenter(context: Any,
                 MessageIdOption(MessageID.male, context, Person.GENDER_MALE),
                 MessageIdOption(MessageID.other, context, Person.GENDER_OTHER))
         view.clazzList = clazzMemberJoinEditHelper.liveList
+        val mWorkSpace = arguments[UstadView.ARG_WORKSPACE]
+        if(mWorkSpace != null){
+            workSpace = Json.parse(WorkSpace.serializer(), mWorkSpace)
+        }
+
+        serverUrl = if (arguments.containsKey(UstadView.ARG_SERVER_URL)) {
```
mikedawson, line 73: I think the serverUrl should be required if it is in registration mode. The serverUrl should be nullable as it would be null if it is not in registration mode.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/PersonEditPresenter.kt
```diff
@@ -2,38 +2,53 @@ package com.ustadmobile.core.controller
 
 import com.ustadmobile.core.db.UmAppDatabase
 import com.ustadmobile.core.generated.locale.MessageID
+import com.ustadmobile.core.impl.AppConfig
```
mikedawson, line 5: I think for this to follow the crud pattern this EditPresenter would also need to use PersonWithAccount (to include the passwords etc) so that the view edits the object directly (instead of the presenter having to ask for other properties).

mikedawson (changes requested):
Generally worked very well. I would request a slight refactor to ensure that this follows the crud patterns. Thanks!

mikedawson (approved):

Issue #123, opened 2020-04-26
Add container upload support [3pts]


mikedawson:
A container that is added on the device (e.g. using the builtin content editor or otherwise imported) must be uploaded to the server.

PR #162, opened 2020-07-22
Dev networkdi upload

SamihMustafa:


sharedse/src/commonMain/kotlin/com/ustadmobile/sharedse/network/containeruploader/UploadJobRunner.kt
```diff
@@ -0,0 +1,143 @@
...
+import kotlinx.atomicfu.atomic
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import org.kodein.di.DI
+import org.kodein.di.DIAware
+import org.kodein.di.instance
+import org.kodein.di.on
+
+class UploadJobRunner(private val containerUploadJob: ContainerUploadJob, private val endpointUrl: String, override val di: DI) : DIAware {
```
mikedawson: If the upload job fails there needs to be a short delay before trying again (e.g. 1second). That should be a parameter to uploadjobrunner (a default can be defined as a constant). To make the tests run faster, you could reduce that delay.

app-android/src/main/assets/about.html
```diff
@@ -1,64 +0,0 @@
-<html>
```
mikedawson: Is there a reason for this being removed in this commit?

app-ktor-server/src/jvmMain/kotlin/com/ustadmobile/lib/rest/ResumableUploadRoute.kt
```diff
@@ -0,0 +1,104 @@
...
+            val sessionId = call.request.header(SESSIONID) ?: ""
+
+            val (isValid, message) = call.request.isValidRequest(folder)
+            if (!isValid) {
+                call.respond(HttpStatusCode.BadRequest, message)
+            } else {
+
+                val input = call.receiveStream()
+                val uploadFile = File(folder, sessionId)
+                val fileOut = FileOutputStream(uploadFile, true)
```
mikedawson: Fileoutputstream is not closed! I would suggest Kotlin's .use extension (which automatically closes something afterwards). It is also important to call flush to ensure everything was written. E.g.
```
FileOutputStream(uploadFile, true).use { fileOut ->
   fileOut.write(input.readBytes())
   fileOut.flush()
}
```

sharedse/src/commonJvmMain/kotlin/com/ustadmobile/sharedse/network/containeruploader/ContainerUploader.kt
```diff
@@ -0,0 +1,169 @@
...
+
+            }
+            downloadStatus
+        }
+
+    }
+
+    companion object {
+
+        const val CHUNK_SIZE = 1024 * 8
```
mikedawson: I would call this DEFAULT_CHUNK_SIZE (as that is what it is), and I would make it 200KB. 8KB is quite small.

sharedse/src/commonJvmMain/kotlin/com/ustadmobile/sharedse/network/containeruploader/ContainerUploaderManagerImp.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
+import com.ustadmobile.core.impl.UstadMobileSystemCommon.Companion.TAG_MAIN_COROUTINE_CONTEXT
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import org.kodein.di.DI
+import org.kodein.di.DIAware
+import org.kodein.di.instance
+import org.kodein.di.on
+
+class ContainerUploaderManagerImp(val endpoint: Endpoint, override val di: DI) : ContainerUploadManager(), DIAware {
```
mikedawson: This is missing an L from the end of the class name

sharedse/src/jvmTest/kotlin/com/ustadmobile/sharedse/network/containeruploader/TestContainerUploader.kt
```diff
@@ -0,0 +1,128 @@
...
+import java.util.concurrent.TimeUnit
+import javax.naming.InitialContext
+
+class TestContainerUploader {
+
+    private lateinit var epubContainer: Container
+    private lateinit var entryListStr: String
+    private lateinit var containerManager: ContainerManager
+    private lateinit var appDb: UmAppDatabase
+    private lateinit var tmpFolder: File
```
mikedawson, line 44: Use the Junit rule so this will be automatically deleted after the test:

```
@JvmField
@Rule
var tmpFolderRule = TemporaryFolder()
```

Then when you need a folder:
```
var tmpFolder = tmpFolderRule.newFolder()
```
... and the rule takes care of deleting it afterwards.

app-ktor-server/src/jvmTest/kotlin/com/ustadmobile/lib/rest/TestResumableUploadRoute.kt
```diff
@@ -0,0 +1,103 @@
...
+import java.nio.charset.Charset
+import kotlin.math.min
+
+class TestResumableUploadRoute {
+
+    lateinit var server: ApplicationEngine
+
+    private val defaultPort = 8098
+
+    lateinit var tmpFolder: File
```
mikedawson, line 38: Please use the Junit temporaryfolder rule here as well.

mikedawson (changes requested):
Generally very good. There are some implementation comments.

mikedawson (approved):

mikedawson:
Implements issue #123

Issue #148, opened 2020-06-28
Update content display screens to record learner progress to ContentEntryProgress table


mikedawson:
Update content display screens to record learner progress to ContentEntryProgress table

PR #166, opened 2020-07-27
Dev crud class progress

SamihMustafa:


app-android/src/main/java/com/ustadmobile/port/android/view/binding/ImageViewBindings.kt
```diff
@@ -110,6 +111,21 @@ fun ImageView.setImageLookupMap(imageLookupMap: Map<Int, Int>?, imageLookupFallb
     updateFromImageLookupMap()
 }
 
+@BindingAdapter(value=["iconStatusFlag"])
+fun ImageView.setIconOnStatusFlag(statusFlag: Int){
```
mikedawson, line 116: This need an else branch. This when clause is used in a recyclerview. You might need to make the statusFlag nullable to make sure that when a view is recycled from an item that has a status flag to one that does not it displays correctly.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/VideoContentPresenterCommon.kt
```diff
@@ -1,6 +1,8 @@
 package com.ustadmobile.core.controller
 
 import com.ustadmobile.core.account.UstadAccountManager
+import com.ustadmobile.core.contentformats.xapi.Statement
```
mikedawson: This class still has the handleUpNavigation() function. Can this be removed as we now use jetpack navigation?

sharedse/src/jvmTest/kotlin/com/ustadmobile/sharedse/util/UstadTestRuleSe.kt
```diff
@@ -74,7 +83,17 @@ class UstadTestRule: TestWatcher() {
...
-            bind<Gson>() with singleton { Gson() }
+            bind<Gson>() with singleton {
+                val builder = GsonBuilder()
+                builder.registerTypeAdapter(Statement::class.java, StatementSerializer())
+                builder.registerTypeAdapter(Statement::class.java, StatementDeserializer())
+                builder.registerTypeAdapter(ContextActivity::class.java, ContextDeserializer())
+                builder.create()
+            }
+
+            bind<XapiStatementEndpoint>() with singleton {
```
mikedawson: Is this really needed for the test rule? Shouldn't any item that is going to use the XapiStatementEndpoint bind it in it's own test (e.g. using a mock).

If it is going to be used, it needs to use the context (e.g. as you have done in UstadApp.kt) used with scoped ... This would connect all xapi statements to one database.

mikedawson:
Please make sure that the statement is only stored if there is an active user: e.g.

```
xapiEndpoint.takeIf { accountManager.activeAccount.personUid != 0L }?.storeStatements(..)
```

mikedawson:
Can you please add a jvm test for the VideoPresenter to test implementation of the updateProgress cases (e.g. when there is an active user and when it is a guest user)?

mikedawson:
Implements most of #148

Issue #182, opened 2020-08-17
Add sort and search to existing list screens


mikedawson:
Add sort and search to existing list screens

Person:
Sort by first name, last name
Search by name

Clazz list:
Sort by name, attendance
Search by name

Clazz Members:
Sort by first name, last name, attendance, date registered, date left
Search by name

Attendance Record:
Search by name
Sort by first name, last name

School Students:
Sort by first name, last name, attendance, date registered, date left
Search by name

School Staff
Sort by first name, last name
Search by name

Class work list:

Sort by deadline (ascending / descending), visible from, name
Search by assignment name

Class work student progress:
Sort by name, sort by content progress, status (e.g. marked, submitted, etc).
Search by student name.

School list:
Sort by name
Search by name

PR #192, opened 2020-09-07
Dev sort search

SamihMustafa:


core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzListPresenter.kt
```diff
@@ -42,13 +32,13 @@ class ClazzListPresenter(context: Any, arguments: Map<String, String>, view: Cla
...
-        view.sortOptions = SortOrder.values().toList().map { ClazzListSortOption(it, context) }
+        selectedSortOption = SORT_OPTIONS[0]
         updateList()
     }
 
-    private fun updateList() {
-        view.list = repo.clazzDao.findClazzesWithPermission(searchQuery,
-                loggedInPersonUid, filterExcludeMembersOfSchool, currentSortOrder.sortOrderCode)
+    private fun updateList(searchText: String? = null) {
+        view.list = repo.clazzDao.findClazzesWithPermission(if(searchText.isNullOrEmpty()) "%%" else "%${searchText}%",
```
mikedawson: I think we should have a string extension function instead of having so many if statements: e.g.

```
fun String?.toQueryLikeParam() = if(isNullOrEmpty()) "%" else "%this%"
```

Then we could use

```
findClazzesWithPermission(searchText.toQueryLikeParam() ...
```

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/PersonDao.kt
```diff
@@ -214,13 +214,15 @@ abstract class PersonDao : BaseDao<Person> {
             AND :timestamp BETWEEN SchoolMember.schoolMemberJoinDate AND SchoolMember.schoolMemberLeftDate )) 
             AND (Person.personUid NOT IN (:excludeSelected))
             AND :accountPersonUid IN ($ENTITY_PERSONS_WITH_SELECT_PERMISSION) 
-            AND Person.firstNames LIKE :searchText
+            AND Person.firstNames || ' ' || Person.lastName LIKE :searchText
```
mikedawson, line 217: Does this query work on postgres?

mikedawson (approved):

mikedawson:
Implements #182

mikedawson (approved):

PR #194, opened 2020-09-17
Apache, Google Drive Importers

SamihMustafa:


app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -184,8 +186,14 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
...
 
+    private fun handleLinkSelection() {
+        onSaveStateToBackStackStateHandle()
+        navigateToPickEntityFromList(ImportedContentEntryMetaData::class.java, R.id.import_link_view)
+    }
+
+
     override suspend fun saveContainerOnExit(entryUid: Long, selectedBaseDir: String, db: UmAppDatabase, repo: UmAppDatabase): Container? {
-        val file = entryMetaData?.file
+        val file = entryMetaData?.uri
```
mikedawson: Why is this variable called file when it could be an http uri?

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -221,6 +229,29 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
...
             entity?.primaryLanguageUid = language.langUid
         }
+
+        navController.currentBackStackEntry?.savedStateHandle?.observeResult(this,
+                ImportedContentEntryMetaData::class.java) {
+            val metadata = it.firstOrNull() ?: return@observeResult
+            // back from navigate import
+            entryMetaData = metadata
+            when (entryMetaData) {
+                null -> {
```
mikedawson: Metadata cannot be null (because it would have returned already). This when statement is therefor redundant. The user should have already received an error on the previous screen if we did not support the link.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -221,6 +229,29 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
...
+                    showSnackBar(getString(R.string.import_link_content_not_supported))
+                }
+                else -> {
+                    selectedUrl = metadata.uri
+                }
+            }
+            val entry = entryMetaData?.contentEntry
+            val entryUid = arguments?.get(ARG_ENTITY_UID)
```
mikedawson, line 232: What is this ? Is this the parent entry uid?

app-ktor-server/src/jvmMain/kotlin/com/ustadmobile/lib/rest/ContentEntryLinkImporter.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.kodein.di.on
+
+@ExperimentalStdlibApi
+fun Route.ContentEntryLinkImporter() {
+
+    route("import") {
+
+        post("validateLink") {
+
+            val url = call.receive<String>()
```
mikedawson: Posting a string as the body seems a bit strange. Post body is normally used for larger payloads. Why not use a normal query parameter for this? (e.g. validateLink?url=...)

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -167,6 +175,21 @@ class ContentEntryEdit2Presenter(context: Any,
 
                         containerUploadManager?.enqueue(uploadJob.cujUid)
                     }
+                }else if (view.selectedUrl != null){
```
mikedawson: There is a logical issue here - because there are two separate variables (selectedUrl and selectedFile). The user could select a file, change their mind, and then select a link.

I think it would be better to have a single variable (e.g. selectedUri). Then the logic can check if selectedUri is a file or an http/https link.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -167,6 +175,21 @@ class ContentEntryEdit2Presenter(context: Any,
...
+
+                    GlobalScope.launch {
```
mikedawson: View.finishWithResult should only be called after the http call has completed.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryImportLinkPresenter.kt
```diff
@@ -0,0 +1,60 @@
...
+        UstadBaseController<ContentEntryImportLinkView>(context, arguments, view, di) {
+
+    val accountManager: UstadAccountManager by instance()
+
+    private val currentHttpClient = defaultHttpClient()
+
+    fun handleClickDone(link: String) {
+        view.showHideProgress(false)
+
+        GlobalScope.launch {
```
mikedawson: Better to use GlobalScope.launch(doorMainDispatcher()) {
  
}

.. then you won't have to have the other two GlobalScope.launch statements.

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/IntExt.kt
```diff
@@ -0,0 +1,6 @@
+package com.ustadmobile.core.util.ext
+
+fun Int.alternative(alternative: Int) = if(this == 0) alternative else this
```
mikedawson, line 3: How is this used? It shows as unused in Android Studio for me.

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/abztract/ScraperTypes.kt
```diff
@@ -1,7 +1,12 @@
 import com.ustadmobile.lib.contentscrapers.abztract.YoutubeChannelIndexer
 import com.ustadmobile.lib.contentscrapers.abztract.YoutubePlaylistIndexer
 import com.ustadmobile.lib.contentscrapers.abztract.YoutubeScraper
+import com.ustadmobile.lib.contentscrapers.apache.ApacheIndexer
```
mikedawson, line 4: There is a spelling error in the package name: ab**z**tract instead of ab**s**tract

Edit - I guess that is on purpose to prevent the use of a keyword...

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/folder/FolderIndexer.kt
```diff
@@ -0,0 +1,85 @@
...
+@ExperimentalStdlibApi
+class FolderIndexer(parentContentEntryUid: Long, runUid: Int, sqiUid: Int, contentEntryUid: Long, endpoint: Endpoint, di: DI) : Indexer(parentContentEntryUid, runUid, sqiUid, contentEntryUid, endpoint, di) {
+
+    override fun indexUrl(sourceUrl: String) {
+
+        try {
+
+            val folder = File(sourceUrl)
+
+            UMLogUtil.logInfo("folder generated ${folder.name}")
```
mikedawson, line 26: We have a UMLogUtil? Why not use Napier directly?

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/abztract/ScraperTypes.kt
SamihMustafa: yes on purpose

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/folder/FolderIndexer.kt
```diff
@@ -0,0 +1,85 @@
...
+
+            fileList.forEach { file ->
+
+                UMLogUtil.logInfo("found file ${file.name}")
+
+                val childEntry = ContentScraperUtil.insertTempContentEntry(contentEntryDao, file.path, folderEntry.primaryLanguageUid, file.name)
+                if (file.isDirectory) {
+                    createQueueItem(file.path, childEntry, ScraperTypes.FOLDER_INDEXER, ScrapeQueueItem.ITEM_TYPE_INDEX, folderEntry.contentEntryUid)
+                } else if (file.isFile) {
+                    createQueueItem(file.path, childEntry, ScraperTypes.FOLDER_SCRAPER, ScrapeQueueItem.ITEM_TYPE_SCRAPE, folderEntry.contentEntryUid)
```
mikedawson, line 75: Why not check to see if the file extension is supported before creating a queue item?

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/folder/FolderScraper.kt
```diff
@@ -0,0 +1,78 @@
...
+                val variant = if (metadataContentEntry.languageVariantUid == 0L)
+                    parentContentEntry?.languageVariantUid ?: contentEntry?.languageVariantUid
+                    ?: 0 else metadataContentEntry.languageVariantUid
+
+                val fileEntry = ContentScraperUtil.createOrUpdateContentEntry(
+                        metadataContentEntry.entryId ?: contentEntry?.entryId ?: name,
+                        metadataContentEntry.title ?: contentEntry?.title ?: name,
+                        sourceUrl, metadataContentEntry.publisher ?: parentContentEntry?.publisher ?: "",
+                        metadataContentEntry.licenseType, primaryLanguage, variant,
+                        metadataContentEntry.description, true, ScraperConstants.EMPTY_STRING,
```
mikedawson: FYI There is no need for an EMPTY_STRING constant - the compiler will automatically do (so there is no performance difference) that and "" is pretty clear.

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/googledrive/GoogleDriveScraper.kt
```diff
@@ -0,0 +1,154 @@
...
+                parameter("fields", "id,modifiedTime,name,mimeType,description,thumbnailLink")
+            }.execute() { fileResponse ->
+
+                val file = fileResponse.receive<GoogleFile>()
+
+                mimeTypeSupported.find { fileMimeType -> fileMimeType == file.mimeType }
+                        ?: return@execute
+
+                val recentContainer = containerDao.getMostRecentContainerForContentEntry(contentEntryUid)
+                val modifiedTime: Long = googleDriveFormat.parse(file.modifiedTime!!).local.unixMillisLong
```
mikedawson: How is modifiedTime nullable? Can we avoid using !! in production code here?

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/folder/FolderIndexer.kt
SamihMustafa: added

lib-content-scrapers/src/jvmMain/kotlin/com/ustadmobile/lib/contentscrapers/googledrive/GoogleDriveScraper.kt
```diff
@@ -0,0 +1,154 @@
...
+                }
+
+                defaultHttpClient().get<HttpStatement>(apiCall) {
+                    parameter("alt", "media")
+                    parameter("key", googleApiKey)
+                }.execute() {
+
+                    val contentFile = File(tempDir, file.name ?: file.id!!)
+                    val stream = it.receive<InputStream>()
+                    FileUtils.writeByteArrayToFile(contentFile, stream.readBytes())
```
mikedawson: this would read the whole file into the memory. It would be better to stream it to disk. Even on the server that is best avoided (e.g. if handling larger video files etc).

lib-content-scrapers/src/jvmTest/kotlin/com/ustadmobile/lib/contentscrapers/folder/TestFolderIndexerAndScraper.kt
```diff
@@ -0,0 +1,119 @@
+package com.ustadmobile.lib.contentscrapers.folder
```
mikedawson: Let's stick to the convention: this should be in two separate files: TestFolderIndexer and TestFolderScraper.

mikedawson (approved):

PR #208, opened 2020-10-06
apache fix

SamihMustafa:


core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/StringExt.kt
```diff
@@ -9,4 +9,8 @@ expect fun String.base64StringToByteArray(): ByteArray
 
 fun String?.toQueryLikeParam() = if(this.isNullOrEmpty()) "%" else "%$this%"
 
+const val forwardSlash = "/"
+
+fun String.requirePostfix(postFix: String = forwardSlash) = if(this.endsWith(postFix)) this else "$this$postFix"
```
mikedawson: Although we are using requirePostfix for adding a / right now, I don't see how "/" is the obvious default value here. e.g. if someone is reading the code and sees someString.requirePostfix(), what the postfix would be by default isn't really straightforward (unlike joinToString, which uses a "," to separate items). 

Can you please remove the default argument here? Sorry to nitpick.

Also can you please use the inline modifier on this?

mikedawson (approved):

Issue #180, opened 2020-08-16
Record H5P Xapi statements


mikedawson:
We should capture H5P Xapi statements and record them when users open H5P content.

mikedawson:
Documentation on : [H5P and XAPI](https://h5p.org/documentation/x-api)

PR #221, opened 2020-10-09
Resolved h5p xapi statement issues

kileha3:
Related to #180 

All statement are sent successfully except for http://adlnet.gov/expapi/verbs/progressed verb statement, it crashed the server (Internal server error) worth looking at.


mikedawson (approved):

mikedawson:
@kileha3  the issue with the statement being submitted is that it had a parent set in contextActivities (under context), but the id of that was null. We need to set that to being the main activity_id.

In the meantime I updated the StatementEndpoint to ignore this for now.

Issue #228, opened 2020-10-11
Edit ContentEntry folder option not visible


mikedawson:
**Describe the bug**
Edit ContentEntry folder option not visible

**To Reproduce**
1. Go to content
2. Create a new folder
3. Open the folder - it cannot be edited

**Expected behavior**
Show an edit option in the extension menu

mikedawson:
In ContentEntryLists:

If listing content by a parent , then show an "Edit" option in the extension menu that starts ContentEntryEdit if the user has edit permission.

PR #237, opened 2020-10-21
add edit option on entryList

SamihMustafa:
#228


app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryList2Fragment.kt
```diff
@@ -119,10 +133,20 @@ class ContentEntryList2Fragment : UstadListViewFragment<ContentEntry, ContentEnt
...
-        if(view?.id == R.id.item_createnew_layout)
+        if (view?.id == R.id.item_createnew_layout)
             mPresenter?.handleClickCreateNewFab()
     }
 
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        when (item.itemId) {
+            R.id.edit -> {
+                mPresenter?.handleEditFolder()
+                return super.onOptionsItemSelected(item)
```
mikedawson: This should be returning true here - the event was handled.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryList2Fragment.kt
```diff
@@ -54,6 +54,12 @@ class ContentEntryList2Fragment : UstadListViewFragment<ContentEntry, ContentEnt
             field = value
         }
 
+    override var hasUpdatePermission: Boolean = false
```
mikedawson: As this is the view let's give it a view like property name (the view doesn't care about underlying permissions etc) e.g. editOptionVisible

mikedawson (approved):

mikedawson:
Resolves issue #228

Issue #235, opened 2020-10-20
Add support for hiding and moving content entry items


mikedawson:
**Is your feature request related to a problem? Please describe.**
Admins need to be able to hide a content item they no longer want others use to see and need to be able to move content from one folder to another.

**Describe the solution you'd like**
Long press in a folder list should result in having options to move or hide items for someone with edit permission on the folder. The extension menu should have options to show or hide hidden content. Hidden content should either have an icon or be semitransparent/faded.

PR #239, opened 2020-10-25
Move or hide Entries

SamihMustafa:
#235


core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryList2Presenter.kt
```diff
@@ -69,13 +83,13 @@ class ContentEntryList2Presenter(context: Any, arguments: Map<String, String>, v
                 Role.PERMISSION_CONTENT_UPDATE)
     }
 
-    private fun getAndSetList(sortOrder: SortOrder = currentSortOrder) {
+    private fun getAndSetList(sortOrder: SortOrder = currentSortOrder, showHidden: Boolean = false) {
```
mikedawson: Why is showHidden a function parameter, and the others are member variables? Let's use the same pattern for all: e.g. showHidden should be a member variable like parentEntryUid, loggedinPersonUid, onlyFolderFilter, etc.

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryList2View.kt
```diff
@@ -12,6 +12,10 @@ interface ContentEntryList2View: UstadListView<ContentEntry, ContentEntryWithPar
      */
     fun showDownloadDialog(args: Map<String, String>)
 
+    fun selectEntry(listOfSelectedEntries: String)
```
mikedawson: Let's be careful with variable naming: this function name does not match what it actually does. The name would make you think that it is going to select an entry on the view. But it is actually going to show the folder list.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryList2Fragment.kt
```diff
@@ -78,20 +84,34 @@ class ContentEntryList2Fragment : UstadListViewFragment<ContentEntry, ContentEnt
...
                 (mDataRecyclerViewAdapter as? ContentEntryListRecyclerAdapter)?.onLocalAvailabilityUpdated(availabilityMap)
             }
         }
+
         setHasOptionsMenu(true)
 
+        var moveCount = navController.currentBackStackEntry?.savedStateHandle?.get<String>(ContentEntryList2View.ARG_MOVING_COUNT)
+        if(!moveCount.isNullOrBlank()){
+            var entryUid = navController.currentBackStackEntry?.savedStateHandle?.get<String>(ARG_CONTENT_ENTRY_UID)
+            showSnackBar("Moved $moveCount entries", {
```
mikedawson: We must avoid any non-localized strings. Strings need to be added to strings_ui.xml

mikedawson (approved):

PR #247, opened 2020-11-17
Dev video encoding

SamihMustafa:


core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryEdit2View.kt
```diff
@@ -22,6 +22,12 @@ interface ContentEntryEdit2View: UstadEditView<ContentEntryWithLanguage>{
 
     var entryMetaData: ImportedContentEntryMetaData?
 
+    var entryCompressed: Boolean
+
+    var videoDimensions: Pair<Int, Int>
+
+    var videoFilePath: String?
```
mikedawson: This is not actually a filepath! It is a URL (as it is also used to represent an http link for video streaming). Can you please rename accordingly?

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryEdit2View.kt
```diff
@@ -22,6 +22,12 @@ interface ContentEntryEdit2View: UstadEditView<ContentEntryWithLanguage>{
 
     var entryMetaData: ImportedContentEntryMetaData?
 
+    var entryCompressed: Boolean
+
+    var videoDimensions: Pair<Int, Int>
```
mikedawson: This is really only a getter. It should be a val with only a getter (no field needs to be initialized).

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryEdit2View.kt
```diff
@@ -22,6 +22,12 @@ interface ContentEntryEdit2View: UstadEditView<ContentEntryWithLanguage>{
 
     var entryMetaData: ImportedContentEntryMetaData?
 
+    var entryCompressed: Boolean
```
mikedawson: 'Compressed' is in the past tense. Nothing at this stage has been compressed or altered yet. Perhaps it can be called compressionEnabled

mikedawson: If this is going to be a var (which makes sense as it is a view property), then it should be on the view binding, and it should use two way binding.

e.g. 

get() = mBinding?.compressionEnabled ?: false
set(value) {
   mBinding?.compressionEnabled = value
}

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/PairExt.kt
```diff
@@ -2,27 +2,33 @@ package com.ustadmobile.core.util.ext
...
+val VALID_RESOLUTIONS = listOf(Pair(480,360), Pair(176,144), Pair(360,480),
+        Pair(144,176), Pair(640, 360), Pair(320, 180), Pair(360, 640), Pair(180, 640),
+        Pair(480,480), Pair(240, 240))
 
 fun Pair<Int, Int>.variance(other: Pair<Int, Int>): Int = abs(this.first - other.first) + abs(this.second - other.second)
 
 fun Pair<Int, Int>.fitWithin(): Pair<Int, Int>{
-    val maxWidth = 640
-    val maxHeight = 360
+    val maxDimension = 640
```
mikedawson: Just to make this function more useful and a bit clearer, how about making maxDimension as a parameter with a default value. E.g. 

```
fun Pair<Int, Int>.fitWithin(maxDimension: Int = 640)
```

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -208,6 +216,10 @@ class ContentEntryEdit2Presenter(context: Any,
                 if (entryUid != null) entry.contentEntryUid = entryUid.toString().toLong()
                 view.fileImportErrorVisible = false
                 view.entity = entry
+                if(metadata.mimeType == "video/mp4"){
```
mikedawson: Why only mp4? Why not just anything that starts with "video/"? This would exclude webm etc.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -86,6 +122,53 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
...
+        player?.prepare(mediaSource)
+    }
+
+    fun prepareWeb(filePath: String){
+        webView?.loadData("""
+            <!DOCTYPE html>
+            <html lang="en">
+            <body>
+            
+             <video id="video" width="100%" height="175px" controls>
```
mikedawson: We don't know for sure that this is an mp4. As there is only one source, you don't need to specify the type, e.g.
```
<video src="$filePath"  controls></video>
```

mikedawson: Also css style attribute should be used for setting width and height.

e.g. style='width: 100%; height: 175px'

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -180,6 +267,24 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
             rootView = it.root
             it.fileImportInfoVisibility = View.GONE
             it.activityEventHandler = this
+            it.entryCompressed = true
+            it.showVideoPreview = false
+            it.showWebPreview = false
+            webView = it.entryEditWebPreview
+            webView?.webChromeClient = WebChromeClient()
+            playerView = it.entryEditVideoPreview
+            webView?.settings?.javaScriptEnabled = true
```
mikedawson, line 276: Why do all these webview setting changes in 'unregisterFileFromTemp()'?

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -8,6 +8,7 @@ import com.ustadmobile.core.networkmanager.defaultHttpClient
 import com.ustadmobile.core.util.MessageIdOption
 import com.ustadmobile.core.util.UMFileUtil
 import com.ustadmobile.core.util.UMUUID
+import com.ustadmobile.core.util.ext.convertToJsonObject
```
mikedawson, line 11: Issue spotted on this presenter that is not directly video related, but related to importing from a url:

handleClickSave's post call to the server needs to be put in a try/catch clause. It is running on the main dispatcher. Any exception will crash the whole app. If there was a connection error or such, the entire app would crash here.

If we get an exception, or a non-200 response status, we should use the snack bar to tell the user that an error happened. In this event, we shouldn't call finish.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -162,6 +169,7 @@ class ContentEntryEdit2Presenter(context: Any,
                             parameter("parentUid", parentEntryUid)
                             parameter("scraperType", view.entryMetaData?.scraperType)
                             parameter("url", view.entryMetaData?.uri)
+                            parameter("conversionParams", Json.stringify(JsonObject.serializer(), conversionParams.convertToJsonObject()))
```
mikedawson: Needs to be split across two lines

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -86,6 +122,53 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
...
+                mBinding?.showWebPreview = true
+                prepareWeb(value)
+            }else{
+                mBinding?.showVideoPreview = true
+                mBinding?.showWebPreview = false
+                prepareVideo(value)
+            }
+        }
+
+    fun prepareVideo(filePath: String) {
```
mikedawson: Naming: prepareVideoFromFile

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -86,6 +122,53 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
...
+        val uri = Uri.parse(filePath)
+        val dataSourceFactory: DataSource.Factory = DefaultDataSourceFactory(requireContext(), "UstadMobile")
+        val mediaSource = ProgressiveMediaSource.Factory(dataSourceFactory)
+                .createMediaSource(uri)
+        player?.prepare(mediaSource)
+    }
+
+    fun prepareWeb(filePath: String){
```
mikedawson: Naming: prepareVideoFromWeb(url: String)

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
mikedawson: This could look prettier by using 
```
webView?.settings?.apply {
  javascriptEnabled = true
  domStorageEnabled = true
}
```

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryEdit2View.kt
SamihMustafa: renamed to VideoUri

app-android/src/androidTest/java/com/ustadmobile/port/android/screen/ContentEntryEditScreen.kt
```diff
@@ -86,18 +93,11 @@ object ContentEntryEditScreen : KScreen<ContentEntryEditScreen>() {
         repo.clientId
         fragmentScenario.clickOptionMenu(R.id.menu_done)
 
-        val entries = dbRule.db.contentEntryDao.findAllLive().waitUntilWithFragmentScenario(fragmentScenario, 15000) {
+        val entries = dbRule.repo.contentEntryDao.findAllLive().waitUntilWithFragmentScenario(fragmentScenario, 15000) {
```
mikedawson: This might not be accurate, and it would get everything out of the database. If something from before had already added any content entry, it would return immediately.

 I would suggest changing this so it looks for the title that was just added. It might also be good to have the title as a parameter of this function.

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryEdit2View.kt
SamihMustafa: done

core/src/commonMain/kotlin/com/ustadmobile/core/view/ContentEntryEdit2View.kt
SamihMustafa: done

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/PairExt.kt
SamihMustafa: done

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
SamihMustafa: done

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
SamihMustafa: done

core/src/commonMain/kotlin/com/ustadmobile/core/networkmanager/ImportJobRunner.kt
```diff
@@ -26,6 +26,10 @@ import kotlinx.atomicfu.atomic
 import kotlinx.coroutines.*
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
+import kotlinx.serialization.builtins.MapSerializer
```
mikedawson, line 29: One issue on importjobrunner's function naming:

There is a function called 'startUpload'. That implies it is async and would continue in the background or something like that. In fact it actually does the whole upload. Can this be renamed to 'doUpload' or just 'upload'?

core/src/jvmMain/kotlin/com/ustadmobile/core/util/ShrinkUtils.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.github.aakira.napier.Napier
+import com.ustadmobile.core.catalog.contenttype.VideoTypePlugin
+import java.io.BufferedReader
+import java.io.File
+import java.io.IOException
+import java.io.InputStreamReader
+
+object ShrinkUtils {
+
+    fun findInPath(commandName: String, systemPropName: String? = null): String? {
```
mikedawson, line 13: This wouldn't work on Windows (as windows commands would have .exe or .bat). I would suggest creating the following extension function for file:

```
/**
 * Checks to see if there is a valid command here. On Linux this is the same as file.exists(). On windows, this will look for the file itself as well as variants with .exe and .bat
 */
fun File.commandExists() {
    return if(linux) {
       exists()
     } else if(windows) }
        File(parent "$name.exe").exists() || File(parent, "$name.bat").exists()
     }
}
```

core/src/jvmMain/kotlin/com/ustadmobile/core/util/ShrinkUtils.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        return System.getenv("PATH").split(File.pathSeparator).firstOrNull() {
+            File(it, commandName).exists()
+        }?.let { File(it, commandName).path }
+    }
+
+    fun getVideoResolutionMetadata(srcFile: File): Triple<Int, Int, String> {
+
+        val ffprobePath = findInPath("ffprobe")
```
mikedawson, line 46: This needs to throw an exception if ffprobe of ffmpeg cannot be found. 

As we are using these for all media conversion I think we should add a warning in UmRestApplication if ffprobe and ffmpeg are not available (so people should see it as soon as they start the server, rather than getting very confused about why media conversion isn't working).

mikedawson (approved):

Issue #232, opened 2020-10-17
Release variant (e.g. obfuscated) crashes when returning from setting class features


mikedawson:
**Describe the bug**
When using the production version of the app (e.g. minified and obfuscated) if the user changes class features, and then returns to class edit, the app will crash.

**To Reproduce**
1. Install the minified (release) version of the app
2. Go to create a class or edit an existing class
3. Change features
4. Select the check box

**Expected behavior**
Enabled features should be updated

**Additional context**
Acrarium bug 13

Issue #202, opened 2020-10-01
Timezones are not populated when assigning a timezone to School/Class


varunasingh:
**Describe the bug**
Timezones are not populated when assigning a timezone to School/Class

**To Reproduce**
1. Go to New School
2. Click on TimeZone
3. Blank.

**Expected behavior**
See timezones and be able to select

**Screenshots**
Self explanatory 

**Desktop (please complete the following information):**
Android 

**Smartphone (please complete the following information):**
Not relevant

**Additional context**
Not relevant

PR #248, opened 2020-11-26
Dev attendance polish

mikedawson:
Improve attendance user experience. Allow users to create schedule a new class (even if there is no set schedule). Fix timezone issue


mikedawson:
Fixes #202

mikedawson:
Fixes #232

PR #250, opened 2020-12-13
Dev report upgrade

SamihMustafa:


core/src/commonMain/kotlin/com/ustadmobile/core/controller/ReportEditPresenter.kt
```diff
@@ -179,6 +152,18 @@ class ReportEditPresenter(context: Any,
...
 
+    fun handleRemoveSeries(){
+
+    }
+
+    fun handleRemoveFilter(){
+
+    }
+
+    fun handleAddSeriesToReport(){
```
mikedawson: fun handleClickAddSeries

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ReportSeriesExt.kt
```diff
@@ -0,0 +1,163 @@
...
+        sql += " , subgroup "
+    }
+
+    sqlList += " GROUP BY StatementEntity.statementUid ORDER BY StatementEntity.timestamp DESC"
+
+
+    return QueryParts(sql, sqlList, paramList.toTypedArray())
+}
+
+private fun groupBy(value: Int): String {
```
mikedawson: Strftime is not compatible with postgres. We need to be able to run all queries on postgres. The query should run on the server when the client is not fully updated.

app-android/src/androidTest/java/com/ustadmobile/port/android/screen/ReportEditScreen.kt
```diff
@@ -1,72 +1,99 @@
...
     val nestScroll = KNestedScrollView {
         withId(R.id.fragment_report_edit_edit_scroll)
     }
 
     fun fillFields(fragmentScenario: FragmentScenario<ReportEditFragment>? = null,
-                   report: ReportWithFilters,
-                   reportOnForm: ReportWithFilters? = ReportWithFilters(),
-                   setFieldsRequiringNavigation: Boolean = true, person: Person? = null,
-                   verbDisplay: VerbDisplay? = null, entry: ContentEntry? = null,
+                   updatedReport: ReportWithSeriesWithFilters,
```
mikedawson, line 88: Let's avoid using the fillFields function for new code now. I think now that we have Kaspresso there it's not really that helpful, and it makes things less clear than just using the normal Kaspresso functions.

mikedawson: @SamihMustafa  can you please also remove fillFields from the screen template?

app-android/src/androidTest/java/com/ustadmobile/port/android/view/DateRangeFragmentTest.kt
```diff
@@ -0,0 +1,156 @@
...
+
+    @Before
+    fun setup() {
+        context = ApplicationProvider.getApplicationContext()
+    }
+
+    @AdbScreenRecord("given Moment not present when filled then should save to database")
+    @Test
+    fun givenNoMomentPresentYet_whenFilledInAndSaveClicked_thenShouldFinishResult() {
+        val fragmentScenario = launchFragmentInContainer(themeResId = R.style.UmTheme_App) {
```
mikedawson: This can go into the init block. You can use a lateinit var if you need to access it later on in the test. 

I think the advantage of using the init block is that it would log where things went wrong (e.g. if there was a crash in the init function).

app-android/src/androidTest/java/com/ustadmobile/port/android/view/DateRangeFragmentTest.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    @Before
+    fun setup() {
+        context = ApplicationProvider.getApplicationContext()
+    }
+
+    @AdbScreenRecord("given Moment not present when filled then should save to database")
+    @Test
+    fun givenNoMomentPresentYet_whenFilledInAndSaveClicked_thenShouldFinishResult() {
+        var fragmentScenario: FragmentScenario<DateRangeFragment>? = null
```
mikedawson: I'd recommend using lateinit here. This is never really going to be null when the test is running

app-android/src/androidTest/java/com/ustadmobile/port/android/view/DateRangeFragmentTest.kt
```diff
@@ -0,0 +1,158 @@
...
+
+
+        }
+    }
+
+
+    @AdbScreenRecord("given Moment exists when updated then should be updated on database")
+    @Test
+    fun givenMomentNotPresentYet_whenValuesForFixedDateNotSet_showErrors() {
+        val fragmentScenario = launchFragmentInContainer(themeResId = R.style.UmTheme_App) {
```
mikedawson: Let's move this into init

app-android/src/androidTest/java/com/ustadmobile/port/android/view/ReportEndToEndTests.kt
```diff
@@ -128,50 +133,27 @@ class ReportEndToEndTests : TestCase() {
...
-                            hasTag("${createdReport.reportUid}")
-                            isDisplayed()
+                    childWith<ReportListScreen.Report> {
+                        withTag(createdReport.reportUid)
+                    } perform {
+                        reportTitle{
+                            hasText(reportToCreate.reportTitle!!)
                         }
                     }
+                    }
```
mikedawson: Indentation here is wrong

app-android/src/androidTest/java/com/ustadmobile/port/android/view/ReportFilterEditFragmentTest.kt
```diff
@@ -0,0 +1,321 @@
...
+    val screenRecordRule = AdbScreenRecordRule()
+
+    @JvmField
+    @Rule
+    var systemImplNavRule = SystemImplTestNavHostRule()
+
+    @AdbScreenRecord("with no report present, fill all the fields and navigate to detail")
+    @Test
+    fun givenNoReportFilterPresentYet_whenPersonGenderFilledIn_thenShouldNavigateBackToReportEditScreen() {
+        val defaultFilter = ReportFilter().apply {
```
mikedawson: Can we move this into init?

app-android/src/androidTest/java/com/ustadmobile/port/android/view/ReportTemplateFragmentTest.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    @JvmField
+    @Rule
+    val screenRecordRule = AdbScreenRecordRule()
+
+
+    @AdbScreenRecord("Given list when ReportTemplate clicked then navigate to ReportEditScreen")
+    @Test
+    fun givenReportTemplateListPresent_whenClickOnReportTemplate_thenShouldNavigateToReportEdit() {
+        val testEntity = Report().apply {
```
mikedawson: Let's put this into init

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzLogListAttendanceFragment.kt
```diff
@@ -189,8 +194,11 @@ class ClazzLogListAttendanceFragment(): UstadListViewFragment<ClazzLog, ClazzLog
                 chart.xAxis.granularity = (1000 * 60 * 60 * 24 * 2).toFloat()
                 chart.axisLeft.axisMinimum = 0f
                 chart.axisLeft.axisMaximum = 100f
-
-                chart.axisLeft.valueFormatter = PercentFormatter(DecimalFormat("###,###,##0"))
+                chart.axisLeft.valueFormatter = object: ValueFormatter(){
```
mikedawson, line 197: Is there any reason to change these on the attendance fragment? The attendance fragment chart is not really customizable like the main report chart.

app-android/src/androidTest/java/com/ustadmobile/test/port/android/util/DateTimeUtils.kt
```diff
@@ -28,10 +28,10 @@ fun setDateField(viewId: Int, timeInMillis: Long, timeZoneId: String = "UTC") {
     KDatePicker {
         withClassName(equalTo(DatePicker::class.java.name))
     } perform {
-        setDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH),
+        setDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH) + 1,
```
mikedawson: Can't we do this as extension function on KView? e.g.

```
fun KView.setDate(timeInMillis: Long, timeZoneId: String) {
   perform {
      click()
   }
   
   val cal = Calendar.getInstance(TimeZone.getTimeZone(timeZoneId))
    cal.timeInMillis = timeInMillis

    KDatePicker {
        withClassName(equalTo(DatePicker::class.java.name))
    } perform {
        setDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH) + 1,
                cal.get(Calendar.DAY_OF_MONTH))

        hasDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH) + 1,
                cal.get(Calendar.DAY_OF_MONTH))
    }
    KView {
        withId(android.R.id.button1)
    } perform {
        click()
    }
}
```

Then it could be used like this:

```
SomeScreen {
    someField {
        setDate(...)
    }
    
}
```

app-android/src/main/java/com/ustadmobile/port/android/impl/ViewNameToDestMap.kt
```diff
@@ -56,9 +56,12 @@ class ViewNameToDestMap: DestinationProvider {
             XapiPackageContentView.VIEW_NAME to UstadDestination(R.id.content_xapi_dest,
                 hideBottomNavigation = true),
             ReportListView.VIEW_NAME to UstadDestination(R.id.report_list_dest),
+            ReportTemplateView.VIEW_NAME to UstadDestination(R.id.report_template_list_dest),
             ReportEditView.VIEW_NAME to UstadDestination(R.id.report_edit_dest,
                     actionBarScrollBehavior = SCROLL_FLAG_NO_SCROLL),
+            ReportFilterEditView.VIEW_NAME to UstadDestination(R.id.report_filter_edit_dest),
```
mikedawson: ReportFilterEditView needs actionBarScrollBehavior = SCROLL_FLAG_NO_SCROLL

app-android/src/main/java/com/ustadmobile/port/android/view/MessageIdAutoCompleteTextView.kt
```diff
@@ -2,13 +2,13 @@ package com.ustadmobile.port.android.view
 
 import android.content.Context
 import android.util.AttributeSet
-import com.ustadmobile.core.util.MessageIdOption
+import com.ustadmobile.core.util.IdOption
 
-class MessageIdAutoCompleteTextView: DropDownListAutoCompleteTextView<MessageIdOption> {
+class MessageIdAutoCompleteTextView: DropDownListAutoCompleteTextView<IdOption> {
```
mikedawson: This class should now be called IdOptionAutoCompleteTextView

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -59,28 +68,49 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
     class ChartViewHolder(val itemBinding: ItemReportChartHeaderBinding) : RecyclerView.ViewHolder(itemBinding.root)
 
     class RecyclerViewChartAdapter(val activityEventHandler: ReportDetailFragmentEventHandler,
-                                   var presenter: ReportDetailPresenter?) : ListAdapter<ReportGraphHelper.ChartData, ChartViewHolder>(DIFFUTIL_CHART) {
+                                   var presenter: ReportDetailPresenter?) : ListAdapter<ChartData, ChartViewHolder>(DIFFUTIL_CHART) {
+
+        var isAdmin = false
+        var chartBinding: ItemReportChartHeaderBinding? = null
```
mikedawson, line 81: Why do we need this chartBinding variable on the recyclerviewadapter? This binding is really part of ChartViewHolder and is unique to each viewholder (even if, in reality, there will only be one viewholder)

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -59,28 +68,49 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
     class ChartViewHolder(val itemBinding: ItemReportChartHeaderBinding) : RecyclerView.ViewHolder(itemBinding.root)
 
     class RecyclerViewChartAdapter(val activityEventHandler: ReportDetailFragmentEventHandler,
-                                   var presenter: ReportDetailPresenter?) : ListAdapter<ReportGraphHelper.ChartData, ChartViewHolder>(DIFFUTIL_CHART) {
+                                   var presenter: ReportDetailPresenter?) : ListAdapter<ChartData, ChartViewHolder>(DIFFUTIL_CHART) {
+
+        var isAdmin = false
```
mikedawson: Here isAdmin is simply about whether or not to show the save as template, correct? In that case, let's call it saveAsTemplateVisible. Deciding who gets permission (and the reasons) is the job of the presenter.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -59,28 +68,49 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
...
+        var chartBinding: ItemReportChartHeaderBinding? = null
+        val boundChartViewHolder = mutableListOf<ChartViewHolder>()
```
mikedawson: boundChartViewHolder is never really used. Values are added, removed, and cleared. But nothing is ever being done with them. Can't this be removed?

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -103,25 +133,78 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
         }
     }
 
+    class AdapterSourceHolder(val statementAdapter: StatementViewRecyclerAdapter,
```
mikedawson, line 138: Can you please add a comment to explain the purpose of AdapterSourceList? We need to explain that we need it because the statement lists is a list of datasources. Each datasource has it's own header, adapter, etc.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -193,32 +275,67 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
...
 
+    override fun onClickExportButton() {
+        val intent = Intent(Intent.ACTION_SEND).apply {
+            flags = Intent.FLAG_ACTIVITY_NEW_TASK
+            addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+            putExtra(Intent.EXTRA_STREAM, getBitmapFromView(chartAdapter?.chartBinding?.
+            previewChartView?.chartView?.chartBitmap))
+            type = "image/png"
+        }
+        startActivity(Intent.createChooser(intent, "Export"))
```
mikedawson: "Export" should not be hardcoded. That should be localized.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -193,32 +275,67 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
...
             findNavController().popBackStack(R.id.report_edit_dest, true)
+            findNavController().popBackStack(R.id.report_template_list_dest, true)
         }
     }
 
+    override fun onClickExportButton() {
+        val intent = Intent(Intent.ACTION_SEND).apply {
+            flags = Intent.FLAG_ACTIVITY_NEW_TASK
+            addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+            putExtra(Intent.EXTRA_STREAM, getBitmapFromView(chartAdapter?.chartBinding?.
```
mikedawson: This is effectively running I/O on the main thread. We shouldn't do that!

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -193,32 +275,67 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
...
+            val file = File(context?.externalCacheDir,
+                    System.currentTimeMillis().toString() + ".png")
+            val out = FileOutputStream(file)
+            bmp?.compress(Bitmap.CompressFormat.PNG, 90, out)
+            out.close()
+            bmpUri = FileProvider.getUriForFile(requireContext(),
+                    "${context?.packageName}.provider", file)
+
+        } catch (e: IOException) {
+            e.printStackTrace()
```
mikedawson, line 313: We should show an error (e.g. SnackBar message) to the user here

app-android/src/main/java/com/ustadmobile/port/android/view/ReportDetailFragment.kt
```diff
@@ -193,32 +275,67 @@ class ReportDetailFragment : UstadDetailFragment<ReportWithFilters>(), ReportDet
...
+            addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+            putExtra(Intent.EXTRA_STREAM, getBitmapFromView(chartAdapter?.chartBinding?.
+            previewChartView?.chartView?.chartBitmap))
+            type = "image/png"
+        }
+        startActivity(Intent.createChooser(intent, "Export"))
+
+    }
+
+    fun getBitmapFromView(bmp: Bitmap?): Uri? {
```
mikedawson: This is not getBitmapFromView. It is saveBitmapAndGetUriForFile

app-android/src/main/java/com/ustadmobile/port/android/view/ReportListFragment.kt
```diff
@@ -49,7 +58,16 @@ class ReportListFragment() : UstadListViewFragment<Report, Report>(),
             holder.itemView.setSelectedIfInList(item, selectedItems, DIFF_CALLBACK)
             (holder.itemBinding.listReportChart.getTag(R.id.tag_graphlookup_key) as? Job)?.cancel()
             val graphJob = GlobalScope.async(Dispatchers.Main) {
-                val chartData = presenter?.getGraphData(item)
+                val series = if(!item.reportSeries.isNullOrEmpty()){
```
mikedawson: I think when we are parsing a JSON we should probably use try/catch. Just in case one chart data gets corrupt somehow (e.g. if we change options etc) we don't want the app to crash or to break other reports.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportTemplateFragment.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.ustadmobile.core.controller.UstadListPresenter
+import com.ustadmobile.core.util.ext.toStringMap
+import com.ustadmobile.core.view.ReportTemplateView
+import com.ustadmobile.lib.db.entities.Report
+import com.ustadmobile.port.android.view.ext.setSelectedIfInList
+import com.ustadmobile.port.android.view.util.ListHeaderRecyclerViewAdapter
+import com.ustadmobile.port.android.view.util.SelectablePagedListAdapter
+
+
+class ReportTemplateFragment(): UstadListViewFragment<Report, Report>(),
```
mikedawson: Should be ReportTemplateListFragment (missing List). Same naming issue goes for presenter, view, tests, etc.

app-android/src/main/java/com/ustadmobile/port/android/view/XapiChartView.kt
```diff
@@ -6,257 +6,241 @@ import android.util.AttributeSet
...
-                    barDataSet.axisDependency = YAxis.AxisDependency.LEFT
-                    barDataSet.color = Color.parseColor(colorList[idx])
-                    barDataSet.setDrawValues(false)
-                    barData.addDataSet(barDataSet)
-                }
-                lineChart.data = barData
-                lineChart.legend.isEnabled = true
 
+        val barData = combinedData.barData ?: BarData()
+        if (barData.dataSetCount > 1) {
```
mikedawson: Can you add a comment to explain why this is happening here?

core/locale/main/values/strings_ui.xml
```diff
@@ -69,6 +69,8 @@
     <string name='classes'>Classes</string>
     <string name='exams'>Exams</string>
     <string name='reports'>Reports</string>
+    <string name="choose_template">Choose template</string>
+    <string name="save_as_template">Save as \ntemplate</string>
```
mikedawson: Why do we need the line break in the template here? Can't we handle that using the wrap options. This will be difficult for translators.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ReportTemplatePresenter.kt
```diff
@@ -0,0 +1,49 @@
...
+                              di: DI, lifecycleOwner: DoorLifecycleOwner)
+    : UstadListPresenter<ReportTemplateView, Report>(context, arguments, view, di, lifecycleOwner) {
+
+    override fun onCreate(savedState: Map<String, String>?) {
+        super.onCreate(savedState)
+        updateListOnView()
+    }
+
+    override suspend fun onCheckAddPermission(account: UmAccount?): Boolean {
+        return true
```
mikedawson: Isn't this always false? There isn't really just a plus button here. The user has to go the report they want, and then click save as a template.

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ReportFilterExt.kt
```diff
@@ -0,0 +1,42 @@
...
+                    selectedOption?.component2() ?: 0, context)
+        }
+        ReportFilter.FIELD_CONTENT_COMPLETION == reportFilterField -> {
+            val selectedOption = ReportFilterEditPresenter.ContentCompletionStatusOption.values()
+                    .find { it.optionVal == reportFilterDropDownValue }
+            valueString = UstadMobileSystemImpl.instance.getString(
+                    selectedOption?.messageId ?: 0, context)
+        }
+        reportFilterCondition == ReportFilter.CONDITION_BETWEEN -> {
+            valueString = """$reportFilterValueBetweenX and $reportFilterValueBetweenY"""
```
mikedawson: and should be localized.

lib-database-entities/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/StatementEntity.kt
```diff
@@ -97,6 +99,11 @@ open class StatementEntity {
 
     var extensionProgress: Int = 0
 
+    /**
+     *  indicates whether or not the statement is about the root contentEntry or child entries
```
mikedawson, line 103: Let's add to the comment:

This is used by queries (e.g. for reports)  e.g. to see if a "completed" verb applies to the contententry itself, or only a subsection (child) of the content

lib-database-entities/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/StatementEntityWithDisplay.kt
```diff
@@ -4,7 +4,7 @@ import androidx.room.Embedded
 import kotlinx.serialization.Serializable
 
 @Serializable
-class StatementListReport : StatementEntity() {
+class StatementEntityWithDisplay : StatementEntity() {
```
mikedawson: Should be StatementEntityWithDisplayDetails  - with display on it's own implies it has a monitor or something

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/ReportFilterDao.kt
```diff
@@ -1,18 +1,14 @@
...
 import com.ustadmobile.lib.db.entities.ReportFilterWithDisplayDetails
 
-@Dao
-@Repository
+//@Dao
+//@Repository
 abstract class ReportFilterDao : BaseDao<ReportFilter>, OneToManyJoinDao<ReportFilter> {
 
-    @Query("""SELECT ReportFilter.*, Person.*, VerbEntity.*, 
+ /*   @Query("""SELECT ReportFilter.*, Person.*, VerbEntity.*,
```
mikedawson: Why do we have commented out code here?

sharedse/src/commonJvmMain/kotlin/com/ustadmobile/port/sharedse/contentformats/xapi/endpoints/XapiUtil.kt
```diff
@@ -261,7 +277,7 @@ object XapiUtil {
                                       instructorUid: Long, agentUid: Long, authorityUid: Long, teamUid: Long,
                                       subActorUid: Long, subVerbUid: Long, subObjectUid: Long,
                                       contentEntryUid: Long = 0L,
-                                      learnerGroupUid: Long): StatementEntity {
+                                      learnerGroupUid: Long, masterRoot: Boolean = false): StatementEntity {
```
mikedawson: masterRoot should be contentEntryRoot

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzLogListAttendanceFragment.kt
SamihMustafa: had to change it when i updated mpandroidchart to 3.1 for combinedChart, this is the new way

mikedawson (approved):

PR #259, opened 2021-02-10
Dev enrolment

SamihMustafa:


app-android/src/androidTest/java/com/ustadmobile/port/android/screen/LeavingReasonEditScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    override val layoutId: Int?
+        get() = R.layout.fragment_report_edit
+    override val viewClass: Class<*>?
+        get() = LeavingReasonEditFragment::class.java
+
+    val LeavingReasonTitleInput = KTextInputLayout { withId(R.id.id_of_textfield_input_layout)}
+
+
+    fun fillFields(fragmentScenario: FragmentScenario<LeavingReasonEditFragment>,
```
mikedawson: Please don't add or use the fillFIelds on any edit screen for testing. As discussed before, now that we have Kaspresso, fillFields causes more problems than it solves.

app-android/src/androidTest/java/com/ustadmobile/port/android/view/ClazzEnrollmentFragmentTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.toughra.ustadmobile.R
+import com.ustadmobile.lib.db.entities.ClazzEnrollment
+import com.ustadmobile.test.rules.SystemImplTestNavHostRule
+import com.ustadmobile.test.rules.UmAppDatabaseAndroidClientRule
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+
+@AdbScreenRecord("ClazzEnrollment screen tests")
+class ClazzEnrollmentFragmentTest : TestCase()  {
```
mikedawson: ClazzEnrollment - What? List, detail?

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEnrollmentEditFragment.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.ustadmobile.core.controller.UstadEditPresenter
+import com.ustadmobile.core.util.ext.observeResult
+import com.ustadmobile.core.util.ext.toStringMap
+import com.ustadmobile.core.view.ClazzEnrollmentEditView
+import com.ustadmobile.lib.db.entities.ClazzEnrollment
+import com.ustadmobile.port.android.util.ext.currentBackStackEntrySavedStateMap
+import com.ustadmobile.port.android.view.ext.navigateToPickEntityFromList
+
+
+interface ClazzEnrollmentFragmentEventHandler {
```
mikedawson: We should refactor to use one 'l' E.g. Enrol as per US english.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzMemberListFragment.kt
```diff
@@ -35,17 +35,17 @@ import com.ustadmobile.port.android.view.util.PagedListSubmitObserver
 import com.ustadmobile.port.android.view.util.PresenterViewLifecycleObserver
 import com.ustadmobile.port.android.view.util.SelectablePagedListAdapter
 
-class ClazzMemberListFragment() : UstadListViewFragment<ClazzMember, ClazzMemberWithPerson>(),
+class ClazzMemberListFragment() : UstadListViewFragment<ClazzEnrollment, ClazzEnrollmentWithPerson>(),
```
mikedawson: Shouldn't this be a list of Person with extra display details (e.g. attendance percentage)?

app-android/src/androidTest/java/com/ustadmobile/port/android/screen/LeavingReasonListScreen.kt
```diff
@@ -0,0 +1,52 @@
...
+    }, itemTypeBuilder = {
+        itemType(::Sort)
+    })
+
+
+    class LeavingReason(parent: Matcher<View>) : KRecyclerItem<LeavingReason>(parent) {
+        val title: KTextView = KTextView(parent) { withId(R.id.line1_text) }
+    }
+
+    class SortOption(parent: Matcher<View>) : KRecyclerItem<SortOption>(parent) {
```
mikedawson: If SortOption is something that we have on every list screen, would it make sense to Create KListScreen and then have those in the parent class?

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEnrollmentListFragment.kt
```diff
@@ -0,0 +1,121 @@
...
+                requireContext().getString(R.string.clazzenrollment))
+
+        mUstadListHeaderRecyclerViewAdapter = ListHeaderRecyclerViewAdapter(
+                this, createNewText)
+        return view
+    }
+
+    override fun onResume() {
+        super.onResume()
+        mActivityWithFab?.activityFloatingActionButton?.text =
```
mikedawson: Is this part of the template? This should be handled by the FabManager, which uses the fragment's lifecycle. This onResume event is the onResume event of the attached activity, not the fragment.

mikedawson (changes requested):
I would recommend doing this in two phases:

1) Do the refactoring of the existing stuff and make sure that everything is still working as expected and tests pass. I think class related stuff that used to link to clazzmember (now clazzenrolment) e.g. ClazzLogAttendanceRecord etc. needs to link directly to personUid. This will avoid the possibility of accidentally showing duplicates and it will make queries simpler (e.g to get the average attendance percentage for a person in a given class, there won't be a need to join enrolment). 

2) Add the new enrolment list screens

mikedawson (approved):

PR #261, opened 2021-02-22
new strings for reason screen

SamihMustafa:


core/locale/main/values/strings_ui.xml
```diff
@@ -507,9 +507,13 @@
     <string name="edit_enrolment">Edit enrolment</string>
     <string name="new_enrolment">New enrolment</string>
     <string name="select_reason">Select reason</string>
+    <string name="select_leaving_reason">Select leaving reason</string>
+    <string name="add_leaving_reason">Add new leaving reason</string>
     <string name="current_status">Current status</string>
     <string name="leaving_reason">Reason for leaving</string>
-    <string name="reason">reason</string>
+    <string name="leaving_reason_add_title">Leaving reason</string>
```
mikedawson: String name should just be "leaving_reason". Doesn't matter if it is used as a title, or something else.

mikedawson: Why do we still have "select reason" if we have "select leaving reason" ?

SamihMustafa: named  it leaving_reason_add_title becuase leaving_reason already exists. removed select reason

mikedawson (approved):

PR #268, opened 2021-03-24
Personedit validations

varunasingh:
Validations on person edit. Awaiting internal testing with latest primary merge.


app-android/src/main/java/com/ustadmobile/port/android/view/PersonEditFragment.kt
```diff
@@ -337,6 +355,37 @@ class PersonEditFragment: UstadEditFragment<PersonWithAccount>(), PersonEditView
             }
         })
 
+        mBinding?.firstnamesText?.addTextChangedListener(object: TextWatcher {
```
mikedawson: To avoid repetition here, let's create a small util class that does this, e.g.

```
class ClearErrorTextWatcher(val textinputlayout: TextInputLayout) {
  ...
  override onTextChanged(...) {
     textinputlayout.isErrorEnabled = false
     textinputlayout.error = null
  }
}
```
Then for each textlayout you can just do:

mBinding?.firstnamesTextinputlayout?.apply {
   mBinding?.firstnamesText?.addTextChangeListener(ClearErrorTextWatcher(it))
}

app-android/src/main/java/com/ustadmobile/port/android/view/PersonEditFragment.kt
```diff
@@ -222,6 +222,24 @@ class PersonEditFragment: UstadEditFragment<PersonWithAccount>(), PersonEditView
             handleInputError(mBinding?.usernameTextinputlayout, value != null, value)
         }
 
+    override var firstNamesFieldError: String? = null
```
mikedawson: These can be handled on the view XML as they are on other views. I think we don't need the handleInputError function.

mikedawson (approved):

PR #269, opened 2021-03-24
attempt list and tabs on contentEntryDetail

SamihMustafa:


lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/ClazzWorkDao.kt
```diff
@@ -192,9 +192,15 @@ abstract class ClazzWorkDao : BaseDao<ClazzWork> {
                 (
                     (
                         SELECT SUM(ContentEntryProgress.contentEntryProgressProgress) 
-                        FROM ContentEntryProgress WHERE
-                        CAST(ContentEntryProgress.contentEntryProgressActive AS INTEGER) = 1
+                        FROM ClazzWorkContentJoin
```
mikedawson, line 196: All new SQL should be spaced according to the SQL style guide

core/src/commonMain/kotlin/com/ustadmobile/core/contentformats/xapi/endpoints/XapiStatementEndpointExt.kt
```diff
@@ -18,6 +19,16 @@ fun XapiStatementEndpoint.storeProgressStatement(account: UmAccount, entry: Cont
             this.id = if (progress == 100) "https://w3id.org/xapi/adl/verbs/satisfied" else "http://adlnet.gov/expapi/verbs/progressed"
             this.display = mapOf("en-US" to if (progress == 100) "satisfied" else "progressed")
         }
+        this.context = XContext().apply {
```
mikedawson, line 22: Why is there a context activity? Isn't this just progressed, where the object is the content piece (e.g. epub, video, etc)? I see why this needs the contextRegistration, but I don't see the need for contextActivities on this statement.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryDetailPresenter.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        val entityJsonStr = bundle[UstadEditView.ARG_ENTITY_JSON]
+        var editEntity: ContentEntry? = null
+        if(entityJsonStr != null) {
+            editEntity = safeParse(di,  ContentEntry.serializer(), entityJsonStr)
+        }else {
+            editEntity = ContentEntry()
+        }
+
+        GlobalScope.launch {
```
mikedawson: There is no need for setupTabs to be suspended. This was done with other presenters where doing a database query was required to see what tabs were going to be displayed. The tab layout here is always the same. The GlobalScope.launch can be removed.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/EpubContentPresenter.kt
```diff
@@ -131,6 +132,8 @@ class EpubContentPresenter(context: Any,
 
     var mCurrentPage: Int = 0
 
+    var contextRegistration: String? = null
```
mikedawson: At the moment we are only using the context registration at the end, in onStop. Let's just have it as a local variable.

For variables that are setup (and never null) in onCreate and then used later on in the lifecycle, we can use lateinit. We know that onCreate will always be called first. We don't need nullability.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/SessionDetailListPresenter.kt
```diff
@@ -0,0 +1,66 @@
...
+    private fun updateListOnView() {
+        view.list = repo.statementDao.findSessionDetailForPerson(contentEntryUid, mLoggedInPersonUid,
+                selectedPersonUid, selectedSession)
+    }
+
+    override fun handleClickCreateNewFab() {
+
+    }
+
+    fun onClickStatementWithSessionDisplay(statementWithStatementDisplay: StatementWithSessionDetailDisplay) {
```
mikedawson: What is the point of this empty function?

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/StatementDao.kt
```diff
@@ -48,7 +48,113 @@ abstract class StatementDao : BaseDao<StatementEntity> {
...
     abstract fun getXLangMap(): XLangMapEntry?
 
+
+    @Query("""UPDATE StatementEntity SET extensionProgress = :progress,
+            statementLastChangedBy = (SELECT nodeClientId FROM SyncNode LIMIT 1) 
+            WHERE statementUid = :uid""")
+    abstract fun updateProgress(uid: Long, progress: Int)
+
+
+    @Query("""
```
mikedawson, line 58: This is not following the SQL style guide.

lib-database-entities/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/PersonWithStatementDisplay.kt
```diff
@@ -0,0 +1,26 @@
+package com.ustadmobile.lib.db.entities
```
mikedawson: This isn't really with "statementdisplay". That implies that it is linked to a specific statement, which it isn't. I would call this PersonWithAttemptsSummary

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryDetailAttemptsListPresenter.kt
```diff
@@ -0,0 +1,84 @@
...
+        updateListOnView()
+    }
+
+
+    override fun handleClickCreateNewFab() {
+
+    }
+
+    fun onClickPersonWithStatementDisplay(personWithStatementDisplay: PersonWithStatementDisplay) {
+            systemImpl.go(SessionsListView.VIEW_NAME,
```
mikedawson: Why are we passing through all the arguments? That could lead to unpredictable behavior. We should only be passing arguments through for specific use cases (e.g. authentication login flow etc).  SessionListView should take ARG_CONTENT_ENTRY_UID and ARG_PERSON_UID.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/EpubContentPresenter.kt
```diff
@@ -142,6 +145,7 @@ class EpubContentPresenter(context: Any,
         super.onCreate(savedState)
         val containerUid = arguments[UstadView.ARG_CONTAINER_UID]?.toLong() ?: 100
         contentEntryUid = arguments[UstadView.ARG_CONTENT_ENTRY_UID]?.toLong() ?: 0
+        contextRegistration = UMUUID.randomUUID().toString()
```
mikedawson: I have removed UMUUID in primary. It was leftover from J2ME days. It's better to use an expect/actual instead. We now have KmpUuid in lib-door-runtime (and the top level function randomUuid() - which is interestingly the example that Kotlin gives for when to use expect/actual - https://kotlinlang.org/docs/mobile/connect-to-platform-specific-apis.html#example-send-and-receive-messages-from-a-websocket ).

core/src/commonMain/kotlin/com/ustadmobile/core/controller/SessionDetailListPresenter.kt
```diff
@@ -0,0 +1,66 @@
...
+    private var selectedPersonUid: Long = 0L
+    private var selectedSession: String = ""
+
+    override suspend fun onCheckAddPermission(account: UmAccount?): Boolean {
+        return false
+    }
+
+    override suspend fun onLoadFromDb() {
+        super.onLoadFromDb()
+        contentEntryUid = arguments[UstadView.ARG_ENTITY_UID]?.toLong() ?: 0
```
mikedawson: ARG_ENTITY_UID is to for when we are passing the uid of a specific entity to load. That isn't really what is happening here (we are actually filtering a list by content entry, person, and session). Please change to using ARG_CONTENT_ENTRY_UID

core/src/commonMain/kotlin/com/ustadmobile/core/controller/SessionsListPresenter.kt
```diff
@@ -0,0 +1,60 @@
+package com.ustadmobile.core.controller
```
mikedawson: Should be SessionListPresenter (we don't use plurals e.g. ClazzListPresenter, PersonListPresenter, etc)

core/src/commonMain/kotlin/com/ustadmobile/core/controller/SessionDetailListPresenter.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.ustadmobile.core.view.UstadView
+import com.ustadmobile.door.DoorLifecycleOwner
+import com.ustadmobile.door.doorMainDispatcher
+import com.ustadmobile.lib.db.entities.StatementWithSessionDetailDisplay
+import com.ustadmobile.lib.db.entities.UmAccount
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import org.kodein.di.DI
+
+class SessionDetailListPresenter(context: Any, arguments: Map<String, String>,
```
mikedawson: Let's call this StatementListPresenter. It is really a list presenter, not a detail presenter.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/VideoContentPresenterCommon.kt
```diff
@@ -60,6 +63,7 @@ abstract class VideoContentPresenterCommon(context: Any, arguments: Map<String,
         containerEntryDao = db.containerEntryDao
         containerDao = db.containerDao
         contentEntryDao = db.contentEntryDao
+        contextRegistration = UMUUID.randomUUID().toString()
```
mikedawson: Please change to using randomUuid from KmpUuid

core/src/commonMain/kotlin/com/ustadmobile/core/controller/VideoContentPresenterCommon.kt
```diff
@@ -38,6 +39,8 @@ abstract class VideoContentPresenterCommon(context: Any, arguments: Map<String,
 
     var timeVideoPlayed = 0L
 
+    var contextRegistration: String? = null
```
mikedawson: Use lateinit instead of nullable. onCreate is always going to be called first. This value will never really be null in the lifecycle

core/src/commonMain/kotlin/com/ustadmobile/core/controller/VideoContentPresenterCommon.kt
```diff
@@ -101,13 +105,15 @@ abstract class VideoContentPresenterCommon(context: Any, arguments: Map<String,
         }
 
         GlobalScope.launch {
+            val registration = contextRegistration ?: UMUUID.randomUUID().toString()
```
mikedawson: This can be removed after changing to lateinit

app-android/src/main/java/com/ustadmobile/port/android/view/binding/ImageViewBindings.kt
```diff
@@ -262,6 +261,36 @@ private fun ImageView.updateFromImageLookupMap() {
...
+        }
+    }else{
+        setImageDrawable(null)
+        context.getString(R.string.incomplete)
+        visibility = View.INVISIBLE
+    }
+}
+
+@BindingAdapter("verbIcon")
+fun ImageView.setVerbIcon(verb: String?) {
```
mikedawson: We already have imageLookupKey and imageLookupMap view bindings (which is a generic handler for setting an image based on some key in a map). It is documented. We don't need an extra view bindings just for verbs. You can use the verb's hardcoded uid constant instead of the url.

mikedawson (changes requested):
Hi - thanks for putting this in. Generally good, please see comments. Please double check that all new SQL strictly follows the SQL style guide (including the 'river').

mikedawson (approved):

PR #279, opened 2021-06-09
Dev clazz content join

SamihMustafa:


mikedawson (changes requested):
Hi,

Can you update this to use Database version 65? This way it can be merged immediately.

core/src/commonMain/kotlin/com/ustadmobile/core/contentformats/xapi/endpoints/XapiStatementEndpointExt.kt
```diff
@@ -29,7 +30,7 @@ fun XapiStatementEndpoint.storeProgressStatement(account: UmAccount, entry: Cont
         }
         this.`object` = XObject().apply {
             this.id = entry.entryId ?: UMFileUtil.joinPaths(account.endpointUrl,
-                    "/contentEntryUid/${entry.contentEntryUid}")
+                    "/contentEntryUid/${entry.contentEntryUid}/$clazzUid/")
```
mikedawson, line 33: @SamihMustafa I think the xObject id should not have clazzUid. It's the same object, regardless of which class it is being used with. This doesn't stop merging it, but can you please remove this for the next pull request?

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryList2Presenter.kt
```diff
@@ -46,26 +46,25 @@ class ContentEntryList2Presenter(context: Any, arguments: Map<String, String>, v
...
+    override val sortOptions: List<SortOrderOption>
+        get() = SORT_OPTIONS
 
     override fun onCreate(savedState: Map<String, String>?) {
         super.onCreate(savedState)
         contentEntryListItemListener.mListMode = mListMode
         contentEntryListItemListener.presenter = this
-        view.sortOptions = SortOrder.values().toList().map { ContentEntryListSortOption(it, context) }
+        selectedSortOption = SORT_OPTIONS[0]
         contentFilter = arguments[ARG_CONTENT_FILTER].toString()
```
mikedawson, line 59: @SamihMustafa  - for the next commit, let's rename this arg. It's not really a filter. It's more like a display by option.  E.g. we have display by class, by parent (e.g. parent content entry),  by downloaded, etc.

mikedawson (approved):

PR #286, opened 2021-06-27
Dev import uri UI

SamihMustafa:


app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryAddOptionsBottomSheetFragment.kt
```diff
@@ -20,45 +32,103 @@ import com.ustadmobile.core.view.UstadView.Companion.ARG_PARENT_ENTRY_UID
...
+            R.id.content_add_link -> {
+                if(isUpdateContent){
+                    navigateToPickEntityFromList(ImportedContentEntryMetaData::class.java,
+                            R.id.import_link_view, args.toBundle())
+                }else{
+                    findNavController().navigate(R.id.import_link_view, args.toBundle())
+                }
+            }
+            R.id.content_add_gallery ->{
+                args[SelectFileView.ARG_SELECT_FILE] = SelectFileView.SELECT_GALLERY
```
mikedawson: This needs to be named appropriately... it's not really a file... it's a selection mode.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryAddOptionsBottomSheetFragment.kt
```diff
@@ -20,45 +32,103 @@ import com.ustadmobile.core.view.UstadView.Companion.ARG_PARENT_ENTRY_UID
...
+
+
+        val args = mutableMapOf(ARG_LEAF to true.toString())
+        args.putFromOtherMapIfPresent(argsMap, ARG_PARENT_ENTRY_UID)
+        args.putFromOtherMapIfPresent(argsMap, KEY_SELECTED_ITEMS)
+
+        when(view?.id){
+
+            R.id.content_add_link -> {
+                if(isUpdateContent){
```
mikedawson: This should use an interface... 

e.g. ContentEntryAddOptionsListener

fun onClickNewFolder

fun onClickImportFile

fun onClickImportLink

Each presenter (e.g. ContentEntryEdit or ContentEntryList) implements that listener interface accordingly.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryAddOptionsBottomSheetFragment.kt
```diff
@@ -20,45 +32,103 @@ import com.ustadmobile.core.view.UstadView.Companion.ARG_PARENT_ENTRY_UID
...
+                }
+            }
+            R.id.content_create_folder ->{
+                args[ARG_LEAF] = false.toString()
+                findNavController().navigate(R.id.content_entry_edit_dest, args.toBundle())
+            }
+        }
+
+
+        if (isShowing != null && isShowing) {
```
mikedawson: This should go... when the user makes a selection, we assume its job is done, and the sheet should be dismissed.

mikedawson: Instead of isShowing - make the listener nullable. Then after the user makes their first selection, set the listener to null (so that repeat events will be handled)

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -316,6 +289,24 @@ class ContentEntryEdit2Fragment(private val registry: ActivityResultRegistry? =
             }
             loading = false
         }
+
+        navController.currentBackStackEntry?.savedStateHandle?.observeResult(this,
```
mikedawson: This is to be moved to the core presenter...

Also: add a key - normally that would be set according to the result type. Because this is just a string, we should give it a keyname so it is clear.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryList2Fragment.kt
```diff
@@ -139,8 +139,16 @@ class ContentEntryList2Fragment : UstadListViewFragment<ContentEntry, ContentEnt
 
     override fun showContentEntryAddOptions(parentEntryUid: Long) {
         val entryAddOption = ContentEntryAddOptionsBottomSheetFragment()
-        entryAddOption.arguments = UMAndroidUtil.mapToBundle(mapOf(UstadView.ARG_PARENT_ENTRY_UID
-                to parentEntryUid.toString()))
+        val args = mutableMapOf(
+                UstadView.ARG_PARENT_ENTRY_UID to parentEntryUid.toString())
+
+        // add selectedItems to args if in stack
+        if(findNavController().previousBackStackEntry?.savedStateHandle?.contains(KEY_SELECTED_ITEMS) == true){
```
mikedawson: Don't look back - pass the arguments forward. This can be done using the multiplatform navigate for results.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -87,6 +96,16 @@ class ContentEntryEdit2Presenter(context: Any,
     override suspend fun onLoadEntityFromDb(db: UmAppDatabase): ContentEntryWithLanguage? {
         val entityUid = arguments[ARG_ENTITY_UID]?.toLong() ?: 0
         val isLeaf = arguments[ARG_LEAF]?.toBoolean()
+        val metaData = arguments[ARG_IMPORTED_METADATA]
+        val uri = arguments[ARG_URI]
+        if(uri != null){
+            handleFileSelection(uri)
+            return null
+        }
+        if(metaData != null){
```
mikedawson: Logic issue for displaying the update button:

Logic is exclusively: show the button when contentEntryUid != 0 and contentEntry is a leaf.

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -87,6 +96,16 @@ class ContentEntryEdit2Presenter(context: Any,
...
+            view.entryMetaData = safeParse(di, ImportedContentEntryMetaData.serializer(), metaData)
```
mikedawson: Other issue here: check the db argument - and run it only when db is Repository. This function will be called twice!!!

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryImportLinkPresenter.kt
```diff
@@ -47,11 +61,23 @@ class ContentEntryImportLinkPresenter(context: Any, arguments: Map<String, Strin
...
+                        args.putFromOtherMapIfPresent(arguments, ARG_PARENT_ENTRY_UID)
+                        systemImpl.go(ContentEntryEdit2View.VIEW_NAME,
+                                args, context)
+                    }
+
 
                 }
 
-            }catch (e: Exception){
+            } catch (e: Exception) {
```
mikedawson, line 74: Here -we should show an error for the user e.g. snackbar

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryList2Presenter.kt
```diff
@@ -178,10 +179,37 @@ class ContentEntryList2Presenter(context: Any, arguments: Map<String, String>, v
         }
     }
 
+    /**
+     * Show ContentEntryList in picker mode so the user can select a folder to move entries to.
+     *
+     * @param selectedContentEntryParentChildJoinUids a string which is a comma separated list of
+     * the selected ContentEntryParentChildJoinUids (e.g. that should be saved to the savedStateHandle)
+     */
+    fun handleShowFolderOnlyEntries(selectedContentEntryParentChildJoinUids: String){
```
mikedawson: handleClickMove(childrenToMove: List<ContentEntryWithEtc>)

There is no longer any need to save KEY_SELECTED_ITEMS because the next presenter should not be looking back...

The move should happen when the selection is made (e.g. in ContentEntrySelectFolder or ContentEntryEdit if creating a new folder) - ensure consistent approach.

mikedawson (changes requested):

mikedawson (approved):

PR #288, opened 2021-07-05
viewpager2

SamihMustafa:


mikedawson (approved):

mikedawson:
This generally looks good, so I'll merge it now. Only one small change to make as an extra (tiny) pull request: 

Instead of using an extension function to create the tab mediator strategy, let's use a child class of TabMediator to do that. I think that's a cleaner, more object-oriented way of doing things.

mikedawson (approved):

PR #296, opened 2021-07-25
Primary datetime timezone binding

varunasingh:
Preparing pull request. 
Whats left: 
removing old references to dateLong


app-android/src/main/java/com/ustadmobile/port/android/view/binding/DatePickerBindingAdapter2.kt
```diff
@@ -0,0 +1,277 @@
...
+        val cal = adapterCalendar
+        val defaultCal = if(!adapterCalendar.timeInMillis.isSet2){
+            Calendar.getInstance()
+        }else{
+            cal
+        }
+
+        val impl : UstadMobileSystemImpl = (context.applicationContext as DIAware
+                                            ).di.direct.instance()
+        val localeString = impl.getLocale(context)
```
mikedawson: This will cause a problem when the phone itself is set to Farsi. impl.getLocale will get the setting. If the user has already set their phone to Farsi, and the app is running in "use system locale", this will return a blank string. 

Use getDisplayedLocale - this will always return the language that is actually on the screen.

app-android/src/main/java/com/ustadmobile/port/android/view/binding/DatePickerBindingAdapter2.kt
```diff
@@ -0,0 +1,277 @@
...
+                .setTitleType(PersianDatePickerDialog.NO_TITLE)
+                .setShowInBottomSheet(true)
+                .setListener(object : PersianPickerListener {
+                    override fun onDateSelected(@NotNull persianPickerDate: PersianPickerDate) {
+                        Toast.makeText(
+                            context,
+                            persianPickerDate.getPersianYear().toString() + "/" +
+                                    persianPickerDate.getPersianMonth() + "/" +
+                                        persianPickerDate.getPersianDay() + " timestamp: " + persianPickerDate.timestamp
+                                        + " converted : " + UMCalendarUtil.getPrettyDateFromLong(persianPickerDate.timestamp, null) ,
```
mikedawson: Toasts should be removed for production usage.

app-android/src/main/java/com/ustadmobile/port/android/view/binding/DatePickerBindingAdapter2.kt
```diff
@@ -0,0 +1,277 @@
...
+@SuppressLint("SetTextI18n")
+@BindingAdapter(value = ["dateTimeMillisFrom", "dateTimeMillisTo"])
+fun TextView.setTextFromToDateTimeMillis(textFromDateLong: Long, textToDateLong: Long) {
+    val impl : UstadMobileSystemImpl = (context.applicationContext as DIAware
+            ).di.direct.instance()
+    val localeString = impl.getLocale(context)
+    val persianDateFrom = PersianDateImpl()
+    persianDateFrom.setDate(textFromDateLong)
+
+    val persianDateTo = PersianDateImpl()
```
mikedawson: Let's move initializing these formatters into the if statement. There is no need to init the Persian calendar when it is not in use (and vice versa)

app-android/src/main/java/com/ustadmobile/port/android/view/binding/DatePickerBindingAdapter2.kt
```diff
@@ -0,0 +1,277 @@
...
+    val calendar = Calendar.getInstance()
+    calendar.timeInMillis = timeInMillis
+
+    init(calendar[Calendar.YEAR], calendar[Calendar.MONTH], calendar[Calendar.DAY_OF_MONTH]) {_, _, _, _ ->
+        bindingListener?.onChange()
+    }
+}
+
+@InverseBindingAdapter(attribute = "timeInMillis")
+fun DatePicker.getTimeInMillis() : Long{
```
mikedawson: This does not apply the time zone, hence it will return the wrong value.

app-android/src/main/java/com/ustadmobile/port/android/view/binding/DatePickerBindingAdapter2.kt
```diff
@@ -0,0 +1,277 @@
...
+            cal
+        }
+
+        val impl : UstadMobileSystemImpl = (context.applicationContext as DIAware
+                                            ).di.direct.instance()
+        val localeString = impl.getLocale(context)
+
+        if(localeString.startsWith("ps") || localeString.startsWith("fa")){
+
+            val picker: PersianDatePickerDialog = PersianDatePickerDialog(context)
```
mikedawson, line 120: This date picker will not be connected to the time zone. This could use the PersianCalendar that comes with the library which will accept a timezone or ( https://github.com/amirmehdizadeh/JalaliCalendar )

mikedawson (approved):

PR #298, opened 2021-08-01
De contentjob

SamihMustafa:


core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentJobRunner.kt
```diff
@@ -0,0 +1,274 @@
...
+                    metadataResult?.pluginId ?: contentPluginManager.extractMetadata(sourceUri,
+                        processContext)?.pluginId ?:
+                        throw FatalContentJobException("ContentJobItem #${item.contentJobItem?.cjiUid}: cannot determine pluginId")
+                }else {
+                    item.contentJobItem?.cjiPluginId ?: 0
+                }
+
+                val plugin = contentPluginManager.getPluginById(pluginId)
+
+                val job = async {
```
mikedawson: I think this should use the deferred result e.g.

```
val jobResult = async {
   plugin.processJob(item, processContext, this@ContentJobRunner)
}
```

then val processResult = jobResult.await()

That's cleaner than doing the assignment inside the async.

mikedawson: I think this should use the deferred result e.g.

```
val jobResult = async {
   plugin.processJob(item, processContext, this@ContentJobRunner)
}
```

then val processResult = jobResult.await()

That's cleaner than doing the assignment inside the async.

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentPluginManagerImpl.kt
```diff
@@ -0,0 +1,23 @@
+package com.ustadmobile.core.contentjob
```
mikedawson: Actually - there are no alternative implementations of this. There is no reason to have a separate implementation and interface. Let's just rename ContentPluginManagerImpl to ContentPluginManager and remove the interface.

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/JobConnectivityLiveData.kt
```diff
@@ -0,0 +1,37 @@
+package com.ustadmobile.core.contentjob
+
+import com.ustadmobile.core.networkmanager.ConnectivityLiveData
+import com.ustadmobile.door.DoorLiveData
+import com.ustadmobile.door.DoorMediatorLiveData
+
+class JobConnectivityLiveData(
```
mikedawson, line 14: Class declaration should use Kotlin style (e.g. ending brackets should be on a new line)

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentJobRunner.kt
SamihMustafa: done but had to keep the var for processResult because in finally block needed to check the status

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentPluginManagerImpl.kt
SamihMustafa: done

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/JobConnectivityLiveData.kt
mikedawson: I think we should use the word mediator in the name of this class because that is it's purpose... e.g.

JobConnectivityLiveDataMediator

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/JobConnectivityLiveData.kt
```diff
@@ -0,0 +1,37 @@
+package com.ustadmobile.core.contentjob
+
+import com.ustadmobile.core.networkmanager.ConnectivityLiveData
+import com.ustadmobile.door.DoorLiveData
+import com.ustadmobile.door.DoorMediatorLiveData
+
```
mikedawson, line 6: Let's add a comment about why this class is here...

e.g.

LiveDataMediator that combines live data for the current connectivity and whether or not the job can run on a metered network. This LiveData can then be used as a source to make a decision about whether or not a job needs to stop.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
```diff
@@ -0,0 +1,121 @@
...
+class UstadTorrentManagerImpl(
+        val endpoint: Endpoint,
+        override val di: DI
+) : UstadTorrentManager, CommunicationManagerListener {
+
+    private val torrentDir: File by di.on(endpoint).instance(tag = DiTag.TAG_TORRENT_DIR)
+
+    private val containerDir by di.on(endpoint).instance<File>(tag = DiTag.TAG_DEFAULT_CONTAINER_DIR)
+
+    private var containerInfoHashMap = mutableMapOf<Long, String>()
```
mikedawson: These multiple maps should be replaced with a single list that uses a data class with the info needed.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/util/ext/ContentPluginCommonJvmExt.kt
```diff
@@ -0,0 +1,47 @@
...
+                }
+
+                db.containerEntryFileDao.deleteListOfEntryFiles(zombieEntryFilesList)
+            } while (zombieEntryFilesList.isNotEmpty())
+        }
+
+        val containers = db.containerDao.findContainersForContentEntryUid(contentEntryUid)
+
+        containers.forEach {
+            ustadTorrentManager.removeTorrent(it.containerUid)
```
mikedawson: Here the actual .torrent file should be removed AFTER the torrent is removed.

app-android/src/androidTest/java/com/ustadmobile/port/android/view/ContentEntryEdit2FragmentTest.kt
```diff
@@ -1,5 +1,6 @@
 package com.ustadmobile.port.android.view
 
+import android.content.ContentResolver
```
mikedawson: This is unused - should be removed.

app-android/src/androidTest/java/com/ustadmobile/port/android/view/ContentEntryEdit2FragmentTest.kt
```diff
@@ -20,6 +21,7 @@ import com.ustadmobile.test.rules.UmAppDatabaseAndroidClientRule
 import org.junit.Assert
 import org.junit.Rule
 import org.junit.Test
+import org.mockito.kotlin.mock
```
mikedawson: this is also unused - should be removed.

app-android/src/androidTest/java/com/ustadmobile/port/android/DownloadJobConnectivityChangeTriggerEventsTest.kt
```diff
@@ -8,10 +8,7 @@ import com.ustadmobile.door.DoorObserver
 import com.ustadmobile.door.entities.NodeIdAndAuth
 import com.ustadmobile.door.ext.DoorTag
 import com.ustadmobile.door.util.randomUuid
-import com.ustadmobile.lib.db.entities.ConnectivityStatus
-import com.ustadmobile.lib.db.entities.ContentEntry
-import com.ustadmobile.lib.db.entities.DownloadJob
-import com.ustadmobile.lib.db.entities.DownloadJobItem
+import com.ustadmobile.lib.db.entities.*
```
mikedawson: I'm not sure what this test is really testing. As far as I can see, it is really only testing livedata. Not sure where this test came from, but it seems pointless. Also it's waiting quite a long time, looks like 10secs+ because of thread.sleep

app-android/src/main/java/com/ustadmobile/port/android/impl/UstadApp.kt
```diff
@@ -99,13 +99,17 @@ open class UstadApp : BaseUstadApp(), DIAware {
 
         bind<UmAppDatabase>(tag = TAG_DB) with scoped(EndpointScope.Default).singleton {
             val dbName = sanitizeDbNameFromUrl(context.url)
-            DatabaseBuilder.databaseBuilder(applicationContext, UmAppDatabase::class, dbName)
-                .addSyncCallback(instance(), false)
+            val nodeIdAndAuth: NodeIdAndAuth = instance()
+            val db = DatabaseBuilder.databaseBuilder(applicationContext, UmAppDatabase::class, dbName)
```
mikedawson: Why do we need val db = ... This assignment and variable can be removed.

app-android/src/main/java/com/ustadmobile/port/android/impl/UstadApp.kt
```diff
@@ -122,6 +126,40 @@ open class UstadApp : BaseUstadApp(), DIAware {
                 it.setupAssignmentSyncListener(context, di)
             }
         }
+        bind<File>(tag = DiTag.TAG_TORRENT_DIR) with scoped(EndpointScope.Default).singleton{
+            val torrentDir = File(applicationContext.filesDir.siteDataSubDir(context), "torrent")
```
mikedawson: There should be a constant for the torrent dir name instead of using a plain string for "torrent"

app-android/src/main/java/com/ustadmobile/port/android/impl/UstadApp.kt
```diff
@@ -122,6 +126,40 @@ open class UstadApp : BaseUstadApp(), DIAware {
...
+            torrentDir.mkdirs()
+            torrentDir
+        }
+
+        bind<ContainerStorageManager> () with scoped(EndpointScope.Default).singleton{
```
mikedawson, line 124: I think we have too much logic going on here in the binding. We could do this using an interface and a different implementation on different platforms, or use expect/actual. Expect/actual can be specified without a constructor function, so each implementation can have it's own constructor args (e.g. context on Android, something else on JVM)

core/src/commonMain/kotlin/com/ustadmobile/core/impl/UMStorageDir.kt
```diff
@@ -39,6 +39,7 @@ import kotlin.jvm.JvmOverloads
  *
  * @author mike
  */
+@Deprecated("Please use ContainerStorageDir via ContainerStorageManager")
 class UMStorageDir
```
mikedawson: Is there anything else still using this?

mikedawson:
Overall note: please make sure that unused imports are removed. Some have unused imports of deprecated classes. These need to be removed, otherwise these unused imports can create a problem for any future refactoring.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -86,7 +89,13 @@ class ContentEntryEdit2Fragment() : UstadEditFragment<ContentEntryWithLanguage>(
         set(value) {
             mBinding?.fileImportInfoVisibility = if (value?.uri != null)
                 View.VISIBLE else View.GONE
-            mBinding?.importedMetadata = value
+            field = value
+        }
+
+    override var metadataResult: MetadataResult? = null
+        set(value) {
+            // TODO need to remove fileSelectedInfo if metadata came from scraper
```
mikedawson, line 87: What is this pending TODO?

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryListRecyclerAdapter.kt
```diff
@@ -79,8 +89,12 @@ class ContentEntryListRecyclerAdapter(var itemListener: ContentEntryListItemList
         holder.itemView.setSelectedIfInList(item, selectedItems, ContentEntryList2Fragment.DIFF_CALLBACK)
         if(item != null) {
             GlobalScope.launch(Dispatchers.Main.immediate) {
-                holder.downloadJobItemLiveData = containerDownloadManager
-                        .getDownloadJobItemByContentEntryUid(item.contentEntryUid )
+                holder.downloadJobItemLiveData = RateLimitedLiveData(appDatabase, listOf("ContentJobItem"), 1000) {
```
mikedawson: There is no need to use GlobalScope.launch here.

mikedawson: This needs to handle when item is null, eg if the item is null, then holder.downloadJobItemLiveData should be set to null.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryListRecyclerAdapter.kt
```diff
@@ -79,8 +89,12 @@ class ContentEntryListRecyclerAdapter(var itemListener: ContentEntryListItemList
...
+                    appDatabase.contentJobItemDao.findStatusForActiveContentJobItem(item.contentEntryUid)
+                }
+                holder.contentJobItemProgress = RateLimitedLiveData(appDatabase, listOf("ContentJobItem"), 1000) {
```
mikedawson: Can't we use an object with embedded fields to do one query instead of two queries? Or would that make things more complicated on the binder or something?

mikedawson:
I think there are some changes in primary that are not yet merged. This includes the onboarding screen (which was updated to use new illustrations).

mikedawson:
I remember there was some mention of complex code for add folder on Android, but I don't see anything complex there. Am I missing something?

app-ktor-server/src/main/kotlin/com/ustadmobile/lib/rest/ContentEntryLinkImporter.kt
```diff
@@ -0,0 +1,111 @@
...
+fun Route.ContentEntryLinkImporter() {
+
+    route("import") {
+
+        post("validateLink") {
+            val url = call.request.queryParameters["url"]?: ""
+            val pluginManager: ContentPluginManager by closestDI().on(call).instance()
+            var metadata: MetadataResult? = null
+            try{
+                val processContext = ProcessContext(createTemporaryDir("content"), mutableMapOf())
```
mikedawson: It looks like this temporary directory is never getting deleted. It should be deleted in a finally block.

app-ktor-server/src/main/kotlin/com/ustadmobile/lib/rest/ContentEntryLinkImporter.kt
```diff
@@ -0,0 +1,111 @@
...
+                val pluginId = call.request.queryParameters["pluginId"]?.toInt() ?: 0
+                val conversionParams = call.request.queryParameters["conversionParams"]
+                Napier.i("contentEntry from client: ${contentEntry.contentEntryUid}")
+
+                val db: UmAppDatabase by closestDI().on(call).instance(tag = DoorTag.TAG_DB)
+                val repo: UmAppDatabase by closestDI().on(call).instance(tag = DoorTag.TAG_REPO)
+                val containerFolder: File by closestDI().on(call).instance(tag = DiTag.TAG_DEFAULT_CONTAINER_DIR)
+                val entryFromDb = db.contentEntryDao.findByUid(contentEntry.contentEntryUid)
+                val endpoint = Endpoint(call.request.header("Host") ?: "localhost")
+                if (entryFromDb == null) {
```
mikedawson, line 81: Why do we need to stop here? This looks like a race condition.

app-ktor-server/src/main/kotlin/com/ustadmobile/lib/rest/StartFile.kt
```diff
@@ -0,0 +1,53 @@
...
+
+fun Route.StartFile(){
+
+    get("/startFile"){
+
+        val di : DI by closestDI()
+        val repo: UmAppDatabase = di.on(call).direct.instance(DoorTag.TAG_REPO)
+        val epubPlugin: EpubTypePluginCommonJvm = di.on(call).direct.instance()
+
+        val epub = File("/home/ustad/server/test.epub")
```
mikedawson: What is this doing?! This looks like testing code of some kind... Shouldn't this be deleted?

app-ktor-server/src/main/kotlin/com/ustadmobile/lib/rest/TorrentFileRoute.kt
```diff
@@ -0,0 +1,108 @@
...
+            call.respondFile(torrentFile)
+        }
+
+        post("{containerUid}/{contentEntryUid}/upload"){
+
+            val torrentDir: File by closestDI().on(call).instance(tag = DiTag.TAG_TORRENT_DIR)
+            val contentJobManager: ContentJobManager by closestDI().instance()
+            val containerFolder: File by closestDI().on(call).instance(tag = DiTag.TAG_DEFAULT_CONTAINER_DIR)
+            val db: UmAppDatabase by closestDI().on(call).instance(tag = UmAppDatabase.TAG_DB)
+            // TODO handle virtualhostmode for endpoint
```
mikedawson: This TODO is a major problem when it is running in a singleton mode. Then nothing will match. KTOR will allow you to access the config, so that you can handle this.

app-ktor-server/src/main/kotlin/com/ustadmobile/lib/rest/TorrentTracker.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.turn.ttorrent.tracker.TrackedTorrent
+import com.turn.ttorrent.tracker.Tracker
+import com.ustadmobile.core.account.Endpoint
+import com.ustadmobile.core.util.DiTag
+import org.kodein.di.*
+import java.io.File
+
+class TorrentTracker(val endpoint: Endpoint, override val di: DI): DIAware {
+
+    private var tracker: Tracker = di.direct.instance()
```
mikedawson: Should be val instead of var

app-ktor-server/src/main/kotlin/com/ustadmobile/lib/rest/TorrentTracker.kt
```diff
@@ -0,0 +1,33 @@
...
+        torrentDir.listFiles()?.forEach {
+            addTorrentFile(it)
+        }
+    }
+
+    fun addTorrentFile(torrentFile: File){
+        tracker.announce(TrackedTorrent.load(torrentFile))
+    }
+
+    fun addTorrentInfoHash(infoHash: ByteArray){
```
mikedawson: If this is not used, let's remove it.

core/src/jvmTest/kotlin/com/ustadmobile/core/contentjob/TestContentJobRunner.kt
```diff
@@ -0,0 +1,201 @@
...
+import org.junit.Rule
+import kotlin.test.Test
+import org.junit.Before
+import org.kodein.di.*
+import org.mockito.kotlin.any
+import org.mockito.kotlin.mock
+import org.junit.Assert
+import org.mockito.kotlin.stub
+
+class TestContentJobRunner {
```
mikedawson, line 28: This needs a test to check the cancellation of a job item due to unacceptable connectivity. That is an important job of the TestContentjobRunner, but it is not covered by these tests.

e.g. givenContentJobItemRunning_whenConnectivityBecomesUnacceptable_thenShouldCancel

You could create another dummy plugin that simply executes a delay function, and then that should be canceled.

core/src/jvmTest/kotlin/com/ustadmobile/core/contentjob/TestContentJobRunner.kt
```diff
@@ -0,0 +1,201 @@
...
+
+            val done = db.contentJobItemDao.isJobDone(2)
+            Assert.assertTrue("Job completed now", done)
+            Assert.assertFalse("Job was not done until connectivity status changed",
+                doneBeforeConnectivityChange)
+        }
+    }
+
+    @Test
+    fun givenJobCreated_whenJobItemFails_thenShouldRetry() {
```
mikedawson, line 173: This test body is missing.

core/src/jvmTest/kotlin/com/ustadmobile/core/contentjob/TestContentJobRunner.kt
```diff
@@ -0,0 +1,201 @@
...
+        val pluginManager: ContentPluginManager by di.onActiveAccount().instance()
+        numTimesToFail = 1
+
+
+
+
+    }
+
+    @Test
+    fun givenJobCreated_whenJobItemFailsAndExceedsAllowableAttempts_thenShouldFail() {
```
mikedawson, line 189: This test body is missing

core/src/androidMain/kotlin/com/ustadmobile/core/catalog/contenttype/FolderIndexerPlugin.kt
```diff
@@ -0,0 +1,165 @@
...
+                            cjiIsLeaf = true
+                            cjiPluginId = metadataResult.pluginId
+                            cjiParentContentEntryUid = contentJobItem.cjiContentEntryUid
+                            cjiConnectivityNeeded = false
+                            cjiStatus = JobStatus.QUEUED
+                            cjiUid = db.contentJobItemDao.insertJobItem(this)
+                        }
+
+                    }else{
+                        println("no metadata found for file ${fileDocument?.name}")
```
mikedawson: This should be using napier, not println

core/src/androidMain/kotlin/com/ustadmobile/core/catalog/contenttype/VideoTypePluginAndroid.kt
```diff
@@ -7,162 +7,275 @@ import android.media.MediaFormat
...
+                }
+
+
+                val container = db.containerDao.findByUid(contentJobItem.cjiContainerUid)
+                        ?: Container().apply {
+                            containerContentEntryUid = contentJobItem.cjiContentEntryUid
+                            cntLastModified = System.currentTimeMillis()
+                            mimeType = supportedMimeTypes.first()
+                            containerUid = repo.containerDao.insertAsync(this)
+                            contentJobItem.cjiContainerUid = containerUid
```
mikedawson: Line 187 is about contentJobItem, so it should come out of the apply block.

core/src/androidMain/kotlin/com/ustadmobile/core/catalog/contenttype/VideoTypePluginAndroid.kt
```diff
@@ -7,162 +7,275 @@ import android.media.MediaFormat
...
+                            .setGranularity(MediaTransformer.GRANULARITY_DEFAULT)
+                            .setVideoFilters(null)
+                            .build()
+                    )
+
+
+                    try {
+                        videoCompleted.await()
+                    } catch (e: Exception) {
+                        Napier.e(tag = VIDEO_ANDROID, throwable = e, message = e.message ?: "")
```
mikedawson: I think if there is an exception here, we need to stop. Why continue creating and uploading a container if the conversion has failed? It should throw a fatal exception (which the runner will catch)

core/src/androidMain/kotlin/com/ustadmobile/core/contentjob/ContentJobManagerAndroid.kt
```diff
@@ -0,0 +1,31 @@
+package com.ustadmobile.core.contentjob
+
+import android.content.Context
+import androidx.work.*
+import com.ustadmobile.core.account.Endpoint
+import com.ustadmobile.door.getFirstValue
```
mikedawson: Unused import

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ContentPluginExt.kt
```diff
@@ -0,0 +1,59 @@
...
+                                                contentJobItem: ContentJobItem,
+                                                progress: ContentJobProgressListener,
+                                                httpClient: HttpClient, torrentFileBytes: ByteArray,
+                                                endpoint: Endpoint){
+    if(contentNeedUpload) {
+        val containerUid = contentJobItem.cjiContainerUid
+        val contentEntryUid = contentJobItem.cjiContentEntryUid
+        val path = UMFileUtil.joinPaths(endpoint.url, "containers/${containerUid}/$contentEntryUid/upload")
+        val torrentFileName = "${containerUid}.torrent"
+        contentJobItem.cjiServerJobId = httpClient.post(path) {
```
mikedawson: Here we have only uploaded the torrent!! This code needs to wait for the server to actually complete it's download. There needs to be an endpoint on the server that provides updates. Server sent events could be a good choice for this. If the user leaves the app during the upload process, the app will get killed, and the server might not succeed at getting the rest of the file.

core/src/androidMain/kotlin/com/ustadmobile/core/catalog/contenttype/VideoTypePluginAndroid.kt
```diff
@@ -7,162 +7,275 @@ import android.media.MediaFormat
...
+
+                val torrentFileBytes = File(torrentDir, "${container.containerUid}.torrent").readBytes()
+                uploadContentIfNeeded(contentNeedUpload, contentJobItem, jobProgress, httpClient, torrentFileBytes, endpoint)
+
+                repo.containerDao.findByUid(container.containerUid)
+
+                return@withContext ProcessResult(JobStatus.COMPLETE)
+            }catch(c: CancellationException){
+
+                withContext(NonCancellable){
```
mikedawson, line 234: This is not canceling the media transformer... This could cause a problem two ways: 1) trying to delete a file whilst the media transformer is still working on it 2) allowing the media transformer to continue doing CPU intensive work for a long time after the user has said they want to cancel.

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ContentPluginExt.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.ktor.client.request.forms.*
+import io.ktor.http.*
+import org.kodein.di.DI
+
+
+expect suspend fun ContentPlugin.withWifiLock(context: Any, block: suspend () -> Unit)
+
+expect suspend fun deleteFilesForContentEntry(contentEntryUid: Long, di: DI, endpoint: Endpoint): Int
+
+suspend fun ContentPlugin.uploadContentIfNeeded(contentNeedUpload: Boolean,
```
mikedawson: This needs to handle cancellation: If the user cancels, we need to try to tell the server to stop

core/src/androidMain/kotlin/com/ustadmobile/core/impl/UstadMobileSystemImpl.kt
```diff
@@ -346,7 +347,7 @@ actual open class UstadMobileSystemImpl : UstadMobileSystemCommon() {
 
         var umDir = File(storageOptions[deviceStorageIndex], contentDirName!!)
         if (!umDir.exists()) umDir.mkdirs()
-        dirList.add(UMStorageDir(umDir.absolutePath,
+        dirList.add(UMStorageDir(umDir.toURI().toString(),
```
mikedawson: This whole function is unused and can be removed, as can getStorageDirsAsync

core/build.gradle
```diff
@@ -109,6 +109,8 @@ kotlin {
                 implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$version_coroutines"
                 implementation "org.jetbrains.kotlinx:kotlinx-serialization-json:$version_kotlinx_serialization"
                 implementation "com.squareup.okhttp3:okhttp:$version_okhttp"
+                implementation "com.github.SamihMustafa:ttorrent:$version_ttorrent"
+                compileOnly 'net.sf:jargs:1.0'
```
mikedawson: Version should have a variable

core/src/commonJvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/DeleteContainerPlugin.kt
```diff
@@ -0,0 +1,86 @@
...
+    val db: UmAppDatabase by di.on(endpoint).instance(tag = DoorTag.TAG_DB)
+
+    private val ustadTorrentManager: UstadTorrentManager = di.direct.instance<UstadTorrentManager>()
+
+    private val torrentDir: File by di.on(endpoint).instance(tag = DiTag.TAG_TORRENT_DIR)
+
+    override val pluginId: Int
+        get() = PLUGIN_ID
+    override val supportedMimeTypes: List<String>
+        get() = TODO("Not yet implemented")
```
mikedawson: Should implement or return an empty list instead of using TODO placeholder

core/src/commonJvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/DeleteContainerPlugin.kt
```diff
@@ -0,0 +1,86 @@
...
+import com.ustadmobile.lib.db.entities.ContentEntryWithLanguage
+import com.ustadmobile.lib.db.entities.ContentJobItemAndContentJob
+import io.ktor.client.*
+import org.kodein.di.DI
+import org.kodein.di.direct
+import org.kodein.di.instance
+import org.kodein.di.on
+import java.io.File
+
+class DeleteContainerPlugin(
```
mikedawson: This is really DeleteContentEnrtyPlugin ... this is working with a content entry now...

core/src/commonJvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/EpubTypePluginCommonJvm.kt
```diff
@@ -1,111 +1,217 @@
...
+                    val haveConnectivityToContinueJob = checkConnectivityToDoJob(db, jobItem)
+                    if (!haveConnectivityToContinueJob) {
+                        return@withContext ProcessResult(JobStatus.QUEUED)
+                    }
+
+
+                    val torrentFileBytes = File(torrentDir, "${container.containerUid}.torrent").readBytes()
+                    uploadContentIfNeeded(contentNeedUpload, contentJobItem, progress, httpClient, torrentFileBytes, endpoint)
+
+                    repo.containerDao.findByUid(container.containerUid)
```
mikedawson: This line is unused... why findByUid and then do nothing?

core/src/commonJvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/XapiTypePluginCommonJvm.kt
```diff
@@ -1,92 +1,173 @@
...
+                val container = db.containerDao.findByUid(contentJobItem.cjiContainerUid)
+                        ?: Container().apply {
+                            containerContentEntryUid = contentJobItem.cjiContentEntryUid
+                            cntLastModified = System.currentTimeMillis()
+                            mimeType = supportedMimeTypes.first()
+                            containerUid = repo.containerDao.insertAsync(this)
+                            contentJobItem.cjiContainerUid = containerUid
+                        }
+
+                db.contentJobItemDao.updateContainer(contentJobItem.cjiUid, container.containerUid)
```
mikedawson: Why update the container just after creating it? That seems pointless. You could use arguments in addEntriesToCotnainerFromZip to make sure the container is updated (if that is not already done by default)

core/src/commonJvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/XapiTypePluginCommonJvm.kt
```diff
@@ -1,92 +1,173 @@
...
+import org.kodein.di.DI
+import org.kodein.di.instance
+import org.kodein.di.direct
+import org.kodein.di.on
+import io.ktor.client.*
 import org.xmlpull.v1.XmlPullParserFactory
+import kotlinx.coroutines.*
+import kotlinx.coroutines.CancellationException
+
+class XapiTypePluginCommonJvm(
```
mikedawson, line 32: All plugins need to handle the scenario of when something was processed, but then connectivity was not ready at the time. They should not process the content again... I think as standard once processing is done then we should set the container uid on the content job. If container uid on the content job item is 0, that means the container needs processed. Non-zero means the processing is already done.

core/src/commonMain/kotlin/com/ustadmobile/core/contentformats/ContentImportManager.kt
```diff
@@ -7,6 +7,7 @@ import com.ustadmobile.lib.db.entities.ContainerImportJob
 /**
  * Manager that handles importing content
  */
+@Deprecated("use ContentJobPlugin to process imports")
```
mikedawson, line 10: Should be removed

core/src/commonMain/kotlin/com/ustadmobile/core/networkmanager/ImportJobRunner.kt
```diff
@@ -23,6 +23,7 @@ import org.kodein.di.DIAware
 import org.kodein.di.instance
 import org.kodein.di.on
 
+@Deprecated("check ContentJobRunner")
```
mikedawson: Should be removed

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentPlugin.kt
```diff
@@ -0,0 +1,39 @@
+package com.ustadmobile.core.contentjob
+
+import com.ustadmobile.door.DoorUri
+import com.ustadmobile.lib.db.entities.*
+import org.kodein.di.DIAware
+
```
mikedawson, line 6: Let's add a comment here as this is a key interface:


/**
Implementations of ContentPlugin manage how content is imported, downloaded, and uploaded.
*/

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentPluginManager.kt
```diff
@@ -0,0 +1 @@
+
```
mikedawson, line 2: Not sure why the diff does not show here... anyhow, let's add a  brief comment for the class

/* 
 * ContentPluginManager is a simple Singleton that maintains a list of all available ContentPlugins. It has convenience functions that iterate over all available plugins.
 */

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2Presenter.kt
```diff
@@ -318,23 +350,19 @@ class ContentEntryEdit2Presenter(context: Any,
...
-                        !metadata.uri.lowercase().startsWith("https://drive.google.com")) {
-                    view.videoUri = uri
-                }
+            val processContext = ProcessContext(createTemporaryDir("content"), mutableMapOf())
+            val metadata = pluginManager.extractMetadata(DoorUri.parse(uri), processContext)
+                    ?: throw IllegalArgumentException("no metadata found")
+            view.metadataResult = metadata
+            val plugin = pluginManager.getPluginById(metadata.pluginId)
+            entry = metadata.entry
+            arguments[ARG_ENTITY_UID]?.let { entry.contentEntryUid = it.toLong() }
```
mikedawson: .let should be used to capture a result - e.g.

```
val something = someOtherVar?.let { it.lowercase() }
```

You could use .also or a normal if for this situation.

core/src/commonMain/kotlin/com/ustadmobile/core/impl/ConnectivityException.kt
```diff
@@ -0,0 +1,7 @@
+package com.ustadmobile.core.impl
+
+import kotlinx.coroutines.CancellationException
+
+class ConnectivityException(val connectivityMessage: String): CancellationException(connectivityMessage) {
```
mikedawson: I think this should be called ConnectivityCancellationException (because it's not a general IO exception, this is specifically thrown when the connectivity is no longer acceptable. We are only using this with contentjob, so it should go into the contentjob package (rather than the general impl package).

core/src/commonMain/kotlin/com/ustadmobile/core/impl/UMStorageDir.kt
mikedawson: Can this be removed?

core/src/commonMain/kotlin/com/ustadmobile/core/io/ext/UmAppDatabaseContainerIoExt.kt
```diff
@@ -3,16 +3,22 @@ package com.ustadmobile.core.io.ext
...
-                                                   recursive: Boolean = true,
+                                                   recursive: Boolean = true, context: Any, di: DI,
                                                    addOptions: ContainerAddOptions)
 
 expect suspend fun UmAppDatabase.addFileToContainer(containerUid: Long, fileUri: DoorUri,
-                                                    pathInContainer: String, addOptions: ContainerAddOptions)
+                                                    pathInContainer: String, context: Any, di: DI, addOptions: ContainerAddOptions)
 
+
+expect suspend fun UmAppDatabase.addTorrentFileFromContainer(containerUid: Long,
```
mikedawson: We're not really adding a torrent file here like the others, I think it would be better to call this writeContainerTorrentFile

core/src/commonMain/kotlin/com/ustadmobile/core/networkmanager/downloadmanager/ContainerDownloadManager.kt
```diff
@@ -11,32 +11,11 @@ import kotlin.js.JsName
  *   Adding DownloadJobs and component DownloadJobItems
  *   Progress tracking including propogating progress up the content child-parent relationship tree
  */
+@Deprecated("use ContentJobPlugins")
```
mikedawson: I think this should be removed.

core/src/commonMain/kotlin/com/ustadmobile/core/networkmanager/ContainerUploadManager.kt
```diff
@@ -16,6 +16,7 @@ data class ContainerUploaderRequest2(val uploadUuid: String,
                                      val siteUrl: String)
 
 
+@Deprecated("use contentJobPlugin")
```
mikedawson: This should be removed.

core/src/commonMain/kotlin/com/ustadmobile/core/torrent/TorrentDownloadListener.kt
```diff
@@ -0,0 +1,8 @@
+package com.ustadmobile.core.torrent
+
+interface TorrentDownloadListener {
+
+    fun onComplete()
+
+    fun onProgress(progress: Int)
```
mikedawson: Please add a comment here for the progress value. What does this represent? Is this from 0-100, or what is it?

core/src/commonMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManager.kt
```diff
@@ -0,0 +1,17 @@
+package com.ustadmobile.core.torrent
+
+import org.kodein.di.DIAware
+
+interface UstadTorrentManager : DIAware {
```
mikedawson: This needs some documentation.

e.g. explanation of what the TorrentManager is and how it fits in, and then a brief bit of documentation on each function.

```
/*
 * UstadTorrentManager manage torrent seeding and downloading. It is endpoint (context) specific. It will communicate with and manage the underlying torrent servers on the platform.
 */ 
```
Please add brief (e.g. 1-2 sentence) comments for each function

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ContentJobItemExt.kt
```diff
@@ -0,0 +1,53 @@
+package com.ustadmobile.core.util.ext
+
+import com.ustadmobile.core.db.JobStatus
+import com.ustadmobile.core.generated.locale.MessageID
+import com.ustadmobile.core.impl.UstadMobileSystemImpl
+import com.ustadmobile.lib.db.entities.ContentJobItem
+
+fun Int.isStatusQueuedOrDownloading() = this >= JobStatus.WAITING_MIN && this < JobStatus.COMPLETE_MIN
```
mikedawson: The int extension functions aren't used anywhere else...  Two things we could do:

1) Make the int extension function private
2) Better: move this logic into the extension function itself, e.g.

```
fun ContentJobItem?.isStatusQueuedOrDownloading() = this?.cjiRecursiveStatus?.let {it >= JobStatus.WAITING_MIN && it < JobStatus.COMPLETE_MIN }  ?: false
```

core/src/commonMain/kotlin/com/ustadmobile/core/util/DiTag.kt
```diff
@@ -12,12 +12,15 @@ object DiTag {
      * The default directory to which ContainerEntryFile data will be saved. This is not necessarily
      * the only directory.
      */
+    @Deprecated("use ContainerStorageManager")
```
mikedawson, line 15: Can this be removed?

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentPluginManagerImpl.kt
mikedawson: Was that change pushed? Looks like it is still in the branch:
https://github.com/UstadMobile/UstadMobile/blob/de-contentjob/core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentPluginManagerImpl.kt

core/src/commonMain/kotlin/com/ustadmobile/core/contentjob/ContentPluginManager.kt
mikedawson: ContentPluginManager should throw an exception when a duplicate id is used. This could be done using an init blcok e.g.
```
val pluginDuplicates = pluginList.filter { plugin -> 
    pluginList.count { it.pluginId == plugin.pluginId } > 1 
}
if(pluginDuplicates.isNotEmpty()) {
    throw IllegalArgumentException("Duplicate pluginIds in: ${pluginDuplicates.joinToString()}")
}
```

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ContentPluginExt.kt
```diff
@@ -0,0 +1,59 @@
...
+            })
+            onUpload { bytesSentTotal, contentLength ->
+                contentJobItem.cjiItemProgress = (contentJobItem.cjiItemTotal / 2) + (((bytesSentTotal / contentLength) * contentJobItem.cjiItemTotal) / 2)
+                progress.onProgress(contentJobItem)
+            }
+        }
+    }
+}
+
+fun checkConnectivityToDoJob(db: UmAppDatabase, jobItemAndContentJob: ContentJobItemAndContentJob): Boolean{
```
mikedawson: This could be a problem... if the user changed meteredNetworkAllowed on the contentjob whilst processing was going on, this would provide the wrong answer.

I think it would be better to create this as an extension function of the database and answer the whole thing as a query.

eg.
```
//Use a query to do the check. This query is only called when connectivity IS needed, so there is no need for the job item id. It's only purpose is to check if the connectivity is acceptable for the job e.g. connectivity == unmetered or (connectivity == metered and meteredNetworkAllowed == true) 
suspend fun UmAppDatabase.isConnectivityAcceptableForJob(jobUid: Long)
```

core/src/jvmMain/kotlin/com/ustadmobile/core/util/FileUtil.kt
```diff
@@ -0,0 +1,8 @@
+package com.ustadmobile.core.util
+
+import com.ustadmobile.door.DoorUri
+import java.nio.file.Files
+
+actual fun createTemporaryDir(prefix: String): DoorUri {
```
mikedawson, line 6: There are a lot of places where temporary files are being created, but not deleted. How about creating a small .use style function for this: e.g.

```
fun File.useAsTempAndDelete(block: (File) -> Unit) {
    try {
        block(this)
    }finally {
        deleteRecursively()
    }
}
```

So then this could be used as :

```
createTemporaryDir("prefix").useAsTempAndDelete { tmpFile -> 
   //use temp directory
}
```

core/src/jvmTest/kotlin/com/ustadmobile/core/torrent/TestTorrentClient.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.ustadmobile.lib.db.entities.ContentJobItemAndContentJob
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import org.junit.Rule
+import org.kodein.di.*
+import java.net.InetAddress
+import java.net.URL
+
+class TestTorrentClient {
```
mikedawson: What is this test doing? Everything is commented out...

core/src/jvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/ApacheIndexerPlugin.kt
```diff
@@ -0,0 +1,163 @@
...
+                            cjiParentContentEntryUid = contentJobItem.cjiContentEntryUid
+                            cjiParentCjiUid = contentJobItem.cjiUid
+                            cjiConnectivityNeeded = false
+                            cjiStatus = JobStatus.QUEUED
+                            cjiUid = db.contentJobItemDao.insertJobItem(this)
+                        }
+
+                    } else {
+
+                        val processContext = ProcessContext(apacheDir, mutableMapOf())
```
mikedawson: Doing this will lead to the resource being downloaded twice, it will be inefficient. Right now these temporary directories are not being cleaned up. 

I think the Indexer should not try to extract metadata. Just check the mime type. If the mime type is unknown, or the mime type matches a supported format, then create a ContentJobItem for it. Then the ContentJobItem itself will download it (once), look at it, and process it (or, ignore it if it can't be processed).

core/src/commonJvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/XapiTypePluginCommonJvm.kt
SamihMustafa: this was updating the contentJobItem containerUid, updated the name

core/src/jvmMain/kotlin/com/ustadmobile/core/catalog/contenttype/ApacheIndexerPlugin.kt
```diff
@@ -0,0 +1,168 @@
...
+import kotlinx.coroutines.withContext
+import org.jsoup.Jsoup
+import org.kodein.di.DI
+import org.kodein.di.direct
+import org.kodein.di.instance
+import org.kodein.di.on
+import java.io.InputStream
+import java.net.URL
+
+class ApacheIndexerPlugin(private var context: Any, private val endpoint: Endpoint, override val di: DI): ContentPlugin {
```
mikedawson, line 23: ApacheIndexer should be removed from lib-content-scrapers

core/src/commonMain/kotlin/com/ustadmobile/core/util/FileUtil.kt
```diff
@@ -0,0 +1,5 @@
+package com.ustadmobile.core.util
+
+import com.ustadmobile.door.DoorUri
+
+expect fun createTemporaryDir(prefix: String = ""): DoorUri
```
mikedawson, line 5: This should be suspended. On JS any temporary directory creation will be async.

core/src/jvmMain/kotlin/com/ustadmobile/core/util/ShrinkUtils.kt
```diff
@@ -81,6 +82,27 @@ object ShrinkUtils {
...
 
 
+    /**
+     * Validate a ratio string that should be in the form of "x:y" where x and y are positive,
+     * non-zero integers. ffprobe might return N/A or something other than a valid aspect ratio,
+     * so this input needs validated.
+     *
+     * @return the ratioStr trimmed if it is valid, null otherwise
+     */
+    fun validateRatio(ratioStr: String): String? {
```
mikedawson, line 92: This should be internal (it is accessed by test functions, but it is not really part of any API.

core/src/commonMain/kotlin/com/ustadmobile/core/util/SymLink.kt
```diff
@@ -0,0 +1,3 @@
+package com.ustadmobile.core.util
+
+expect fun createSymLink(oldPath: String, targetPath: String)
```
mikedawson: This expect/actual can only be used by JVM and Android. I think the expect signature should be in commonJvmMain

core/src/jvmTest/kotlin/com/ustadmobile/core/catalog/contenttype/ContentDispatcher.kt
```diff
@@ -0,0 +1,48 @@
...
+                response.setHeader("Content-Type", mimetype)
+            }
+            if (!request.method.equals("HEAD", ignoreCase = true))
+                response.setBody(buffer)
+
+            return response
+
+        } catch (e: Exception) {
+            e.printStackTrace()
+            System.err.println(request.path)
```
mikedawson, line 42: Should return a 500 internal error response here.

mikedawson:
1) All ContentPlugin tests need to test handling invalid input
2) ContentPlugins that have long running process operations (e.g. video) need to test handling cancellation 
3) The the default content uploader plugin should have a test where the upload logic is tested (e.g. using communicationmanager).

core/src/jvmTest/kotlin/com/ustadmobile/core/contentformats/GlobalDisptacher.kt
```diff
@@ -0,0 +1,63 @@
...
+import okio.Okio
+import okio.buffer
+import okio.source
+import org.apache.commons.io.IOUtils
+import java.io.File
+import java.io.IOException
+import java.nio.file.Files
+
+
+val globalDisptacher: Dispatcher = object : Dispatcher() {
```
mikedawson, line 16: This "global" dispatcher is used in only one test, and it is really the same concept as ContentDispatcher.

I think you need one class: ResourcesDispatcher. That could take in a base path argument e.g.

```
ResourcesDispatcher(private val resourceBasePath: String)

 fun dispatch(request) {
    val resPath = UMFileUtil.joinPaths(resourceBasePath, request.path)
  }
}
```

core/src/jvmTest/kotlin/com/ustadmobile/core/controller/ContentEntryEdit2PresenterTest.kt
```diff
@@ -1,17 +1,19 @@
+/*
```
mikedawson: Why is this entire test commented out?

core/src/jvmTest/kotlin/com/ustadmobile/core/controller/ContentEntryImportLinkPresenterTest.kt
```diff
@@ -5,7 +5,9 @@ import org.mockito.kotlin.mock
 import org.mockito.kotlin.timeout
 import org.mockito.kotlin.verify
 import com.ustadmobile.core.account.UstadAccountManager
+import com.ustadmobile.core.catalog.contenttype.EpubTypePluginCommonJvm
```
mikedawson, line 5: import com.ustadmobile.core.contentformats.metadata.ImportedContentEntryMetaData is now redundant and should be removed

core/src/jvmTest/kotlin/com/ustadmobile/core/io/UploadSessionTest.kt
```diff
@@ -64,9 +64,6 @@ class UploadSessionTest {
         Napier.baseDebugIfNotEnabled()
```
mikedawson: Can't we remove uploadsession and uploadsessionmanager? These are not being used anymore.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    private var containerListenerMap = mutableMapOf<Long, DownloadListenerAdapter>()*/
+
+    // on android, di uses connectionManager and a provider to get the current communicationManager
+    // on jvm, communicationManager is a singleton
+    private val communicationManager: UstadCommunicationManager?
+        get() = di.direct.instanceOrNull()
+
+    // lock add and removeTorrent
+    private val torrentLock = Mutex()
```
mikedawson: Adding is locked only by things that call addTorrent (eg startSeeding). The lock should be called within addTorrent itself.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    private val containerDir by di.on(endpoint).instance<File>(tag = DiTag.TAG_DEFAULT_CONTAINER_DIR)
+
+    private var containerInfoHashMap = mutableMapOf<Long, String>()
+
+    private var activeTorrentDownloadMap = mutableMapOf<Long, ActiveTorrentInfo>()
+
+ /*   private var containerDownloadPathMap = mutableMapOf<Long, String>()
+
+    private var containerManagerMap = mutableMapOf<Long, TorrentManager>()
```
mikedawson, line 41: Should be val

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    private val torrentDir: File by di.on(endpoint).instance(tag = DiTag.TAG_TORRENT_DIR)
+
+    private val containerDir by di.on(endpoint).instance<File>(tag = DiTag.TAG_DEFAULT_CONTAINER_DIR)
+
+    private var containerInfoHashMap = mutableMapOf<Long, String>()
+
+    private var activeTorrentDownloadMap = mutableMapOf<Long, ActiveTorrentInfo>()
+
+ /*   private var containerDownloadPathMap = mutableMapOf<Long, String>()
```
mikedawson: These commented out items should be removed

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
mikedawson: These maps should be val instead of var.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    // on android, di uses connectionManager and a provider to get the current communicationManager
+    // on jvm, communicationManager is a singleton
+    private val communicationManager: UstadCommunicationManager?
+        get() = di.direct.instanceOrNull()
+
+    // lock add and removeTorrent
+    private val torrentLock = Mutex()
+
+    val pieceStorage: FairPieceStorageFactory by lazy {
```
mikedawson: Should be private

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
```diff
@@ -0,0 +1,138 @@
...
+        withContext(Dispatchers.Default){
+
+            val startTime = System.currentTimeMillis()
+
+            val torrentFile = File(torrentDir, "$containerUid.torrent")
+
+            val containerFilesDir = if(downloadPath != null) File(downloadPath)
+                                    else File(containerDir, containerUid.toString())
+            containerFilesDir.mkdirs()
+            if(!torrentFile.exists())
```
mikedawson: I think this a little bit of a mine in the code. There is the documentation or naming that would indicate to anyone looking at this public function that they need to make the torrent first. I think the code that creates a new torrent file should be put here.

Because this is the "Torrent Manager", it seems like a more logical place for this code to live, as opposed to the download job.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/UstadTorrentManagerImpl.kt
```diff
@@ -0,0 +1,138 @@
...
+        torrentDir.listFiles() { file: File ->
+            file.name.endsWith(".torrent")
+        }?.forEach { torrentFile ->
+            torrentLock.withLock {
+                    addTorrent(torrentFile.nameWithoutExtension.toLong(), null)
+            }
+        }
+    }
+
+    override suspend fun addTorrent(containerUid: Long, downloadPath: String?) {
```
mikedawson: One problem with addTorrent and the active map: I don't see anything that is watching for completion. Once the download is completed, it is no longer an "active" torrent that should be added (even if connectivity is metered). 

I would suggest something like this:

```
inner class TorrentCompleteListener(private val containerUid: Long): TorrentListener {
     ... //blank implementations of all functions needed
    fun onDownloadComplete() {
          //Remove this from the active list
    }

}
```

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/ContainerTorrentDownloadJob.kt
```diff
@@ -0,0 +1,218 @@
...
+                        newContainerEntry.add(containerEntry)
+                    }
+                }
+                db.containerEntryDao.insertListAsync(newContainerEntry)
+
+                contentJobItem.cjiItemProgress = contentJobItem.cjiItemTotal
+                progress.onProgress(contentJobItem)
+
+            } catch (c: CancellationException) {
+                throw c
```
mikedawson: If the job is cancelled, we must remove the torrent from the torrentmanager! Otherwise the download will still continue.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/ContainerTorrentDownloadJob.kt
```diff
@@ -0,0 +1,218 @@
...
+                            println("downloaded torrent in ${System.currentTimeMillis() - startTime} ms")
+                            torrentDeferred.complete(true)
+                        }
+
+                        override fun onProgress(progressSize: Int) {
+                            contentJobItem.cjiItemProgress = (progressSize * 0.75).toLong()
+                            progress.onProgress(contentJobItem)
+                        }
+                    }
+                    GlobalScope.launch(Dispatchers.Default) {
```
mikedawson: This is creating a leak! We should use launch without GlobalScope. It should be in the same scope, so that if the context is canceled, this should also be canceled. The isActive in this loop refers to the new GlobalScope context, so it will always be true. 

This code is actually not needed. Await will throw the cancellation exception if it is canceled.

Then (as per comment below) we need to make sure that when the cancellation exception is thrown, we remove the torrent. 

Perhaps UstadTorrentManager's removeTorrent function should take a parameter to indicate whether or not to delete the actual torrent file. Then the job can call removeTorrent and indicate that it should not delete the torrent file itself.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/ContainerTorrentDownloadJob.kt
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                }
+
+                withWifiLock(context) {
+
+                    val startTime = System.currentTimeMillis()
+                    ustadTorrentManager.addTorrent(containerUid, containerFilesFolder.path)
+
+                    val torrentDeferred = CompletableDeferred<Boolean>()
+                    val torrentListener = object : TorrentDownloadListener {
```
mikedawson: This listener does not handle the failure case! Right now, in case of failure, it will get stuck forever. When the failure event is received, that is the time to call torrentDeferred.completeExceptionally.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/ContainerTorrentDownloadJob.kt
```diff
@@ -0,0 +1,218 @@
...
+                val manifestJson = manifestFile.readText()
+                val manifest: ContainerManifest = Json.decodeFromString(
+                        ContainerManifest.serializer(), manifestJson)
+                val newContainerEntry = mutableListOf<ContainerEntry>()
+                manifest.entryMap?.entries?.forEach { entry ->
+                    if (entry.key == MANIFEST_FILE_NAME) {
+                        return@forEach
+                    }
+                    entry.value.forEach { path ->
+                        val cefUid = db.containerEntryFileDao.findEntryByMd5Sum(entry.key.base64EncodedToHexString())?.cefUid
```
mikedawson: Any idea how long this function takes? Maybe make one query that would select just cefUid and md5sums needed into one map.

core/src/commonJvmMain/kotlin/com/ustadmobile/core/torrent/DownloadListenerAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+    override fun downloadComplete() {
+        destListener.onComplete()
+    }
+
+    override fun pieceReceived(pieceInformation: PieceInformation?, peerInformation: PeerInformation?) {
+
+    }
+
+    override fun downloadFailed(cause: Throwable?) {
+
```
mikedawson: We need to pass this on as well.

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/ContentEntryDao.kt
```diff
@@ -161,6 +109,19 @@ abstract class ContentEntryDao : BaseDao<ContentEntry> {
     @JsName("findByTitle")
     abstract fun findByTitle(title: String): DoorLiveData<ContentEntry?>
 
+    @Query("""
+       SELECT COALESCE((SELECT cjIsMeteredAllowed
```
mikedawson, line 149: This subquery needs a LIMIT 1. In reality, there never really should be more than one job for a given content entry, but better to have it just in case, if there was some split second when one job is stopping (eg download) and another began.

sharedse/src/commonMain/kotlin/com/ustadmobile/sharedse/controller/DownloadDialogPresenter.kt
```diff
@@ -229,42 +196,79 @@ class DownloadDialogPresenter(
...
+            cjIsMeteredAllowed = !isWifiOnlyChecked
+            cjNotificationTitle = impl.getString(MessageID.downloading_content, context)
+                    .replace("%1\$s",entry?.title ?: "")
+            cjUid = appDatabase.contentJobDao.insertAsync(this)
+        }
+        currentJobId = job.cjUid
+
+        ContentJobItem().apply {
+            cjiJobUid = job.cjUid
+            sourceUri = "" // TODO entry ?
```
mikedawson: Set this as it is used with extract metadata

sharedse/src/commonMain/kotlin/com/ustadmobile/sharedse/controller/DownloadDialogPresenter.kt
```diff
@@ -229,42 +196,79 @@ class DownloadDialogPresenter(
...
+            cjiItemTotal = container?.fileSize ?: 0
+            cjiPluginId = 10 // TODO containerTorrentDownload plugin ref in core commonMain
```
mikedawson: TODO here: yes - this should be using the constant, not a magic number!

sharedse/src/commonMain/kotlin/com/ustadmobile/sharedse/controller/DownloadDialogPresenter.kt
```diff
@@ -229,42 +196,79 @@ class DownloadDialogPresenter(
...
-            currentDownloadJobItem.isStatusCompletedSuccessfully() && currentDownloadJobItemVal != null ->
-                createDeleteJobAndRequestPreparation(currentDownloadJobItemVal.djiUid)
-
-            currentDownloadJobItem.isStatusPaused() && currentDownloadJobItemVal != null -> GlobalScope.launch {
-                containerDownloadManager.enqueue(currentDownloadJobItemVal.djiDjUid)
+        when (currentContentJobItemStatus) {
+            ContentJobItem.STATUS_COMPLETE -> GlobalScope.launch {
+                createDeleteJob()
+            }
+            JobStatus.PAUSED -> GlobalScope.launch {
```
mikedawson: Let's just remove this branch as we are not supporting pause right now

mikedawson (changes requested):
As per comments

mikedawson:
Can you please check if you have merged the latest primary?

app-ktor-server/src/main/kotlin/com/ustadmobile/lib/rest/ContentEntryLinkImporter.kt
SamihMustafa: stop where?

mikedawson (approved):

PR #309, opened 2022-01-04
check and retry uploads

SamihMustafa:


core/src/commonJvmMain/kotlin/com/ustadmobile/core/network/containeruploader/ContainerUploader2.kt
```diff
@@ -100,31 +101,38 @@ actual class ContainerUploader2 actual constructor(
...
             pipeIn?.copyTo(NullOutputStream())
             pipeIn?.close()
             response?.closeQuietly()
+        }
 
-            try {
-                httpClient.get<Unit>("${endpoint.url}ContainerUpload2/${request.uploadUuid}/close")
-            }catch(e: Exception){
-                //do nothing
+        try {
```
mikedawson: This can be moved inside the first try/catch - no need for a second one

SamihMustafa: done

mikedawson (approved):

PR #313, opened 2022-01-12
assignment file

SamihMustafa:


core/src/commonMain/kotlin/com/ustadmobile/core/view/SelectFileView.kt
```diff
@@ -12,11 +12,20 @@ interface SelectFileView : UstadView {
 
         const val SELECTION_MODE_GALLERY = "video/*;audio/*"
 
+        const val SELECTION_MODE_VIDEO = "video/*"
```
mikedawson, line 13: It would be better to have a single argument that provides the actual list of mime types. In the case of selecting a file for a content entry, this should come from the contentpluginmanager to get a list of all supported types (instead of duplicating this in a constant).

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzAssignmentDetailOverviewPresenter.kt
```diff
@@ -74,11 +95,43 @@ class ClazzAssignmentDetailOverviewPresenter(context: Any,
...
+                        }
+
+            }
+            db.clazzAssignmentDao.getStatementScoreProgressForAssignment(
                     clazzAssignment.caUid, loggedInPersonUid)
+                    .observeWithLifecycleOwner(lifecycleOwner){
+                        view.clazzMetrics = it
+                    }
+        }else{
+            // isTeacher
```
mikedawson, line 130: Might not be a teacher... just they are not a student. Could be admin etc.

mikedawson:
New entities should not have @MasterChangeSeqNum and @LocalChangeSeqNum

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzAssignmentEditPresenter.kt
```diff
@@ -136,20 +266,32 @@ class ClazzAssignmentEditPresenter(context: Any,
                 return@launch
             }
 
+
             if (entity.caLateSubmissionType == ClazzAssignment.ASSIGNMENT_LATE_SUBMISSION_ACCEPT ||
                     entity.caLateSubmissionType == ClazzAssignment.ASSIGNMENT_LATE_SUBMISSION_PENALTY) {
 
-                if (entity.caDeadlineDate == Long.MAX_VALUE) {
+                if (entity.caDeadlineDate == Long.MAX_VALUE || view.deadlineTime == 0L) {
```
mikedawson, line 273: Let's show all errors for the user as soon as they click submit (e.g. so they don't fix one thing. click again, then see the next error)

core/src/commonMain/kotlin/com/ustadmobile/core/controller/ClazzAssignmentDetailStudentProgressOverviewListPresenter.kt
```diff
@@ -36,6 +36,8 @@ class ClazzAssignmentDetailStudentProgressOverviewListPresenter(context: Any, ar
         mLoggedInPersonUid = accountManager.activeAccount.personUid
         updateListOnView()
         presenterScope.launch {
+            val entity = repo.clazzAssignmentDao.findByUidAsync(clazzAssignmentUid)
```
mikedawson, line 39: This file should be ClazzAssignmentSubmissionListPresenter

mikedawson:
Obsolete

PR #323, opened 2022-04-07
Dev chat message integration

varunasingh:
This is a Work In Progress.
Outstanding work for integration: 

Migrations

Discussions and Topics

Group chats


app-android/src/main/java/com/ustadmobile/port/android/view/ChatListFragment.kt
```diff
@@ -0,0 +1,157 @@
...
+            : DiffUtil.ItemCallback<ChatWithLatestMessageAndCount>() {
+            override fun areItemsTheSame(oldItem: ChatWithLatestMessageAndCount,
+                                         newItem: ChatWithLatestMessageAndCount): Boolean {
+
+                return oldItem.chatUid == newItem.chatUid
+            }
+
+            override fun areContentsTheSame(oldItem: ChatWithLatestMessageAndCount,
+                                            newItem: ChatWithLatestMessageAndCount): Boolean {
+                return oldItem == newItem
```
mikedawson: The diffutil for a read-only item should only compare fields that are displayed to the user. That is important to prevent all the existing chat message viewholders being redrawn needlessly when a new chat message lands.

app-android/src/main/java/com/ustadmobile/port/android/view/binding/TextViewBindings.kt
```diff
@@ -42,6 +48,26 @@ fun TextView.setCustomFieldHint(customField: CustomField?) {
     }
 }
 
+@BindingAdapter("chatMessage", "loggedInPersonUid")
+fun TextView.setChatMessageTitle(message: MessageWithPerson, loggedInPersonUid: Long){
+    if(message.messagePerson?.personUid == loggedInPersonUid){
+        text = systemImpl.getString(MessageID.you, context)
+        gravity = Gravity.RIGHT
```
mikedawson: Shouldn't we use START and END so this works with right to left?

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/MessageDao.kt
```diff
@@ -0,0 +1,76 @@
...
+        WHERE Message.messageTableId = :tableId 
+          AND Message.messageEntityUid = :entityUid
+        ORDER BY Message.messageTimestamp DESC
+    """)
+    abstract fun findAllMessagesByChatUid(entityUid: Long, tableId: Int):
+            DoorDataSourceFactory<Int, MessageWithPerson>
+
+
+    @Insert
+    abstract suspend fun updateAsync(entity: Message)
```
mikedawson: Why is this called updateAsync when it is annotated @Insert ?

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/ChatMemberDao.kt
```diff
@@ -0,0 +1,89 @@
...
+             SET chatMemberPending = true
+          */               
+    """)
+    @ReplicationRunOnChange([ChatMember::class])
+    @ReplicationCheckPendingNotificationsFor([ChatMember::class])
+    abstract suspend fun replicateOnChange()
+
+
+    @Insert
+    abstract suspend fun updateAsync(entity: ChatMember)
```
mikedawson: Why is this called updateAsync when it is annotated @Insert ?

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/dao/ChatDao.kt
```diff
@@ -0,0 +1,177 @@
...
+
+
+    @Query("""
+        SELECT CASE
+                   WHEN Chat.isChatGroup THEN Chat.chatTitle
+                   ELSE Person.firstNames||' '||Person.lastName
+               END AS title
+        FROM Chat
+        LEFT JOIN Person ON Person.personUid =
+          (SELECT pp.personUid
```
mikedawson, line 161: The person wouldn't need to be joined if this is a group. E.g.
LEFT JOIN Person 
          ON CAST(Chat.isChatGroup AS INTEGER) =1 
          AND Person.personUid = ...

app-android/src/main/java/com/ustadmobile/port/android/view/MessagesRecyclerAdapter.kt
```diff
@@ -0,0 +1,97 @@
...
+        val DIFF_CALLBACK_COMMENTS =
+                object : DiffUtil.ItemCallback<MessageWithPerson>() {
+                    override fun areItemsTheSame(oldItem: MessageWithPerson,
+                                                 newItem: MessageWithPerson): Boolean {
+                        return oldItem.messageUid == newItem.messageUid
+                    }
+
+                    override fun areContentsTheSame(oldItem: MessageWithPerson,
+                                                    newItem: MessageWithPerson): Boolean {
+                        return oldItem.messageSenderPersonUid == newItem.messageSenderPersonUid &&
```
mikedawson, line 91: I think we should check the person name here as well. That is what is displayed. The name wouldn't change very often, but it is part of what is shown to the user.

mikedawson (approved):

PR #328, opened 2022-05-09
Moved handle Click logic to impl, added link support in topic detail screen

varunasingh:


app-android/src/main/java/com/ustadmobile/port/android/view/DiscussionPostDescriptionRecyclerAdapter.kt
```diff
@@ -1,45 +1,98 @@
...
+                this.setOnLinkClickListener(onClickListener)
+                this.setOnLinkLongClickListener(onLongClickListener)
+
+            }
+
+        Linkify.addLinks(textView, Linkify.ALL)
+
+    }
+
+    val onClickListener: BetterLinkMovementMethod.OnLinkClickListener =
```
mikedawson: Why not have a class that does this? e.g.

```
class BetterLinkMovementLinkClickListener(val systemImpl: UstadMobileSystemImpl)
```

mikedawson: e.g. class could then be used by all components that need to listen (instead of repeating the same anonymous listeners)?

core/src/androidMain/kotlin/com/ustadmobile/core/impl/UstadMobileSystemImpl.kt
```diff
@@ -485,6 +485,20 @@ actual open class UstadMobileSystemImpl : UstadMobileSystemCommon() {
         (context as Context).startActivity(intent)
     }
 
+    actual fun handleClickLink(url: String, context: Any){
```
mikedawson: This code should be in common. The logic will be the same.

Needs jvm tests.

Needs to handle edge case: when link is an ustad link, but a different endpoint

app-android/src/main/java/com/ustadmobile/port/android/view/BetterLinkMovementClickListener.kt
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+    private val onLongClickListener: BetterLinkMovementMethod.OnLinkLongClickListener =
+        BetterLinkMovementMethod.OnLinkLongClickListener{
+                _,_ ->
+            true
+        }
+
+
+    fun addMovement(textView: TextView?){
```
mikedawson, line 30: Why should the parameter be nullable?

app-android/src/main/java/com/ustadmobile/port/android/view/MessagesRecyclerAdapter.kt
```diff
@@ -47,7 +52,9 @@ class MessagesRecyclerAdapter(
         holder.binding.message = message
 
         holder.binding.itemCommentsListLine2Text.text = message?.messageText
-        addMovement(holder.binding.itemCommentsListLine2Text)
+        val listener = BetterLinkMovementLinkClickListener(systemImpl, accountManager, context)
+        listener.addMovement(holder.binding.itemCommentsListLine2Text)
+        //addMovement(holder.binding.itemCommentsListLine2Text)
```
mikedawson: Commented out line

mikedawson (approved):

PR #377, opened 2022-07-29
Dev 1.7update webfix

kileha3:
I did not migrate to functional components due to a reason mentioned [here](https://github.com/UstadMobile/UstadMobile/blob/dev-1.7update-webfix/app-react/src/main/kotlin/com/ustadmobile/mui/ext/RBuilderExt.kt) . I upgraded all react js wrappers to the latest and did some refactoring on UI elements.


mikedawson:
Has effectively been merged via 1.7 update branch, thanks!

PR #413, opened 2022-10-26
Dev MVVM

ParimSsme:
merge dev-mvvm branch into dev-mvvm-kmp


app-android/src/main/java/com/ustadmobile/port/android/view/OnboardingActivity.kt
```diff
@@ -0,0 +1,391 @@
...
+            .size(size.dp)
+            .clip(CircleShape)
+            .background(gray)
+    )
+    Spacer(modifier = Modifier.width(5.dp))
+}
+
+@Composable
+private fun bottomRow(onClickNext: () -> Unit = { }){
+    val context = LocalContext.current
```
mikedawson: Please replace using context with stringResource

app-android/src/main/res/values/strings.xml
```diff
@@ -3,4 +3,15 @@
 
     <!-- TODO: Remove or change this placeholder text -->
     <string name="hello_blank_fragment">Hello blank fragment</string>
+    <string name="title_activity_text_input">TextInput</string>
```
mikedawson: I think all these Strings are already present in the strings xml. Please remove these duplicates. If we have duplicates that can make extra work for translators.

core/src/androidMain/kotlin/com/ustadmobile/core/shared/OnBoardingModel.kt
```diff
@@ -0,0 +1,13 @@
+package com.ustadmobile.core.shared
```
mikedawson: I think this file isn't used anymore. Is that true? If it is unused, please delete it.

mikedawson:
This has been merged into dev-mvvm-primary

PR #416, opened 2022-11-07
PersonDetail2 screen completed

ParimSsme:


ParimSsme:
created login and siteEnterLink screens with jetpack compose

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -119,5 +146,102 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
+            .verticalScroll(rememberScrollState()),
+        horizontalAlignment = Alignment.CenterHorizontally
+    )  {
+
+        UstadTextEditField(
+            value = "",
```
mikedawson: Value must use uiState - e.g. loginUiState.username

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -119,5 +146,102 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+            label = stringResource(id = R.string.username),
+            onValueChange = {
+                onUsernameValueChange(it)
+            },
+            error = uiState.usernameError,
+            enabled = true,
+        )
+
+        UstadTextEditField(
+            value = "",
```
mikedawson: Value needs to use ui state e.g. uiState.password

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -119,5 +146,102 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+            Text(stringResource(R.string.login).uppercase())
+        }
 
+        Spacer(modifier = Modifier.height(10.dp))
+
+        OutlinedButton(
+            onClick = onClickCreateAccount,
+            modifier = Modifier
+                .fillMaxWidth(),
+            colors = ButtonDefaults.buttonColors(
```
mikedawson: I think there is no need to set the colors here.

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -119,5 +146,102 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+            Text(stringResource(R.string.create_account).uppercase())
+        }
+
+        Spacer(modifier = Modifier.height(10.dp))
+
+        OutlinedButton(
+            onClick = onClickConnectAsGuest,
+            modifier = Modifier
+                .fillMaxWidth(),
+            colors = ButtonDefaults.buttonColors(
```
mikedawson: Again -  I think no need to set the colors here. I think it is already the default for the outlined button.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/LoginViewModel.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.ustadmobile.door.ext.DoorTag
+import com.ustadmobile.lib.db.entities.*
+import kotlinx.coroutines.launch
+import org.kodein.di.on
+
+data class LoginUiState(
+    val username: String = "",
+    val password: String = "",
+    val isEmptyPassword: Boolean = false,
+    val usernameError: String = "",
```
mikedawson: The errors (username error, password error) should be nullable, and should be null by default. Any non-null value means there is an error, and will make the field show red as an error.

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -102,7 +117,19 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+            )
+
+            setContent {
+                MdcTheme {
+                    LoginScreen(uiState)
+                }
+            }
+        }
+
+//        return rootView
```
mikedawson: Please remove any commented out code.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonAccountEditFragment.kt
```diff
@@ -127,7 +149,18 @@ class PersonAccountEditFragment: UstadEditFragment<PersonWithAccount>(), PersonA
...
+                ViewCompositionStrategy.DisposeOnLifecycleDestroyed(viewLifecycleOwner)
+            )
+
+            setContent {
+                MdcTheme {
+                    PersonAccountEditScreen(uiState)
+                }
+            }
+        }
+//        return rootView
```
mikedawson: Please remove commented out code.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/PersonDetailViewModel.kt
```diff
@@ -60,5 +64,10 @@ class PersonDetailViewModel(
         }
     }
 
+    fun getGender(gender: Int = 0): String {
```
mikedawson: I think this should be removed. Is anything still using it? It does not follow the design patterns.

ParimSsme: no, I will remove that

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -119,5 +142,94 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+        )
+
+        Spacer(modifier = Modifier.height(10.dp))
+
+        Button(
+            onClick = onClickLogin,
+            modifier = Modifier
+                .fillMaxWidth(),
+            colors = ButtonDefaults.buttonColors(
+                contentColor = Color.Black,
```
mikedawson: We should never use hardcoded colors! What if the secondary color is also black? If there is a particular background, use contentColorFor

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/PersonAccountEditViewModel.kt
```diff
@@ -0,0 +1,19 @@
...
+    val newPassword: String = "",
+
+    val passwordConfirmed: String = "",
+)
+
+data class PersonAccountEditUiState(
+
+    val personUsernameAndPassword: PersonUsernameAndPasswordModel = PersonUsernameAndPasswordModel(),
+
+    val currentPasswordError: String = "",
```
mikedawson: All error strings must be nullable. Any non-null value, even "", means the component will show in an error state.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/PersonAccountEditViewModel.kt
```diff
@@ -0,0 +1,19 @@
...
+)
```
mikedawson, line 33: Please add a boolean for fieldsEnabled. We should use this on all screens where the user is editing something. FieldsEnabled will normally be true, but will be set to false when loading, saving etc.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonAccountEditFragment.kt
```diff
@@ -178,4 +208,89 @@ class PersonAccountEditFragment: UstadEditFragment<PersonWithAccount>(), PersonA
...
+
+        UstadTextEditField(
+            value = uiState.personUsernameAndPassword.username,
+            label = stringResource(id = R.string.username),
+            onValueChange = {
+                uiState.personUsernameAndPassword.copy(
+                    username = it
+                )
+            },
+            error = "",
```
mikedawson: All fields need to use the UiState for the error (if any) and enabled values.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment.kt
```diff
@@ -198,3 +216,101 @@ class PersonDetailFragment: UstadDetailFragment<PersonWithPersonParentJoin>(), P
 
 }
 
+@Composable
```
mikedawson: We are not using this - everything is in PersonDetailFragment2. I think this Composable should be deleted.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment2.kt
```diff
@@ -79,31 +98,368 @@ class PersonDetailFragment2 : Fragment(){
...
 
 @Composable
 @Preview
 fun PersonDetailScreenPreview() {
-    PersonDetailScreen(
-        person = Person().apply {
-            firstNames = "Bob"
-        }
-    )
+    val uiState = PersonDetailUiState()
```
mikedawson: Please add some data for the preview so it shows more - e.g. add name, phone number, email address, etc.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/LoginViewModel.kt
```diff
@@ -0,0 +1,17 @@
+package com.ustadmobile.core.viewmodel
+
+data class LoginUiState(
+    val username: String = "",
+    val password: String = "",
+    val usernameEnabled: Boolean = true,
```
mikedawson: We should have only one fieldsEnabled property. If username textbox is disabled, so password will always be disabled etc.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/LoginViewModel.kt
```diff
@@ -0,0 +1,17 @@
...
+    val passwordEnabled: Boolean = true,
+    val isEmptyPassword: Boolean = false,
```
mikedawson: I know the field was on the view interface, but please remove the passwordEmpty and isEmptyUsername. Just keep usernameError and passwordError.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/LoginViewModel.kt
```diff
@@ -0,0 +1,17 @@
...
+    val usernameError: String? = null,
+    val passwordError: String? = null,
+    val versionInfo: String = "v42",
+    val isEmptyUsername: Boolean = false,
+    val inProgress: Boolean = false,
```
mikedawson: We don't need the inProgress property on Jetpack compose. We can just keep the fieldsEnabled property.

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -119,5 +142,94 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+    onPasswordValueChange: (String) -> Unit = {},
+) {
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
+            .verticalScroll(rememberScrollState()),
+        horizontalAlignment = Alignment.CenterHorizontally
+    )  {
+
```
mikedawson, line 161: Login Intent message (if not null) should be displayed here above the text fields.

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
```diff
@@ -119,5 +142,94 @@ class Login2Fragment : UstadBaseFragment(), Login2View {
...
+            enabled = uiState.passwordEnabled,
+            trailingIcon = {
+                Icon(
+                    painter = painterResource(R.drawable.ic_baseline_visibility_24),
+                    contentDescription = "")
+            }
+        )
+
+        Spacer(modifier = Modifier.height(10.dp))
+
```
mikedawson, line 186: The main error message should be displayed between the fields and the buttons. This is where invalid username/password, or connection error message would appear.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/PersonAccountEditViewModel.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    val currentPassword: String = "",
+
+    val newPassword: String = "",
+
+    val passwordConfirmed: String = "",
+)
+
+data class PersonAccountEditUiState(
+
```
mikedawson, line 15: The UiState needs to have a nullable error message for each field e.g. usernameError, currentPasswordError, newPasswordError, passwordConfirmedError

It also needs to have a nullable field called "errorMessage" (e.g. in case the app cannot connect, or there is some error that does not relate directly only to one field).

It needs to have currentPasswordVisible and usernameVisible Boolean as per the view interface.

Please add only the fields mentioned above. Don't add other fields from the view interface.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment2.kt
```diff
@@ -79,31 +98,368 @@ class PersonDetailFragment2 : Fragment(){
...
+    onClickChangePassword: () -> Unit = {},
+    onClickManageParentalConsent: () -> Unit = {},
+    onClickChat: () -> Unit = {},
 ) {
-    Text(person?.firstNames ?: "No Name")
+    Row(
+        modifier = Modifier.horizontalScroll(rememberScrollState())
+    ) {
+
+        if (!uiState.person?.phoneNum.isNullOrEmpty()){
```
mikedawson: These properties should be put on the UiState. Then they are shared between Javascript and Android. I have pushed an update for the UiState.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment2.kt
```diff
@@ -79,31 +98,368 @@ class PersonDetailFragment2 : Fragment(){
...
+        Spacer(modifier = Modifier.height(15.dp))
+
+        Button(onClick = onClickClazz) {
+            ClassItem(clazz)
+        }
+    }
+}
+
+@Composable
+private fun ClassItem(clazz: ClazzEnrolmentWithClazzAndAttendance){
```
mikedawson: This needs to show the same information that was shown using the Xml - e.g. list of classes.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment2.kt
```diff
@@ -79,31 +98,368 @@ class PersonDetailFragment2 : Fragment(){
...
+    ){
+        Image(
+            painter = painterResource(id = R.drawable.ic_group_black_24dp),
+            contentDescription = null,
+            colorFilter = ColorFilter.tint(color = Color.Gray),
+            modifier = Modifier
+                .width(70.dp))
+
+        Column {
+            Text(text = "Par")
```
mikedawson: Why is this par text hardcoded?

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment2.kt
```diff
@@ -79,31 +98,368 @@ class PersonDetailFragment2 : Fragment(){
...
+        val gender = messageIdMapResource(
+            map = PersonConstants.GENDER_MESSAGE_ID_MAP,
+            key = uiState.person?.gender ?: 1)
+
+        val dateOfBirth = remember { DateFormat.getDateFormat(context)
+            .format(Date(uiState.person?.dateOfBirth ?: 0)).toString() }
+
+        if (uiState.person?.dateOfBirth != 0L
+            && uiState.person?.dateOfBirth != null){
+            DetailFeild(
```
mikedawson: All use of DetailFeild should be changed to use UstadDetailField.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment2.kt
```diff
@@ -79,31 +98,368 @@ class PersonDetailFragment2 : Fragment(){
...
+                R.drawable.ic_date_range_black_24dp,
+                dateOfBirth,
+                stringResource(R.string.birthday))
+        }
+
+        if (uiState.person?.gender != 0
```
mikedawson: Gender is a mandatory field, so there is no need for an if statement.

app-android/src/main/java/com/ustadmobile/port/android/view/PersonEditFragment.kt
```diff
@@ -173,94 +77,161 @@ class PersonEditFragment: UstadEditFragment<PersonWithAccount>(), PersonEditView
...
+                return PersonEditViewModel(di, SavedStateHandleAdapter(handle)) as T
+            }
+        }
     }
+}
 
-    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
-        super.onViewCreated(view, savedInstanceState)
-        mPresenter?.onCreate(backStackSavedState)
+@Composable
```
mikedawson, line 117: Needs to add username field, password field, parental email field (as we had on the XML layout: https://github.com/UstadMobile/UstadMobile/blob/primary/app-android/src/main/res/layout/fragment_person_edit.xml#L399 ). You can add boolean fields on UiState for them to control if they are visible or not.

app-android/src/main/java/com/ustadmobile/port/android/view/SiteEnterLinkFragment.kt
```diff
@@ -93,4 +126,81 @@ class SiteEnterLinkFragment : UstadBaseFragment(), SiteEnterLinkView{
...
+        Image(
+            painter = painterResource(id = R.drawable.illustration_connect),
+            contentDescription = null,
+            modifier = Modifier
+                .height(200.dp))
+
+        Text(stringResource(R.string.please_enter_the_linK))
+
+        UstadTextEditField(
+            value = "",
```
mikedawson: The value here must come from the view model.

app-android/src/main/java/com/ustadmobile/port/android/view/SiteEnterLinkFragment.kt
```diff
@@ -93,4 +126,81 @@ class SiteEnterLinkFragment : UstadBaseFragment(), SiteEnterLinkView{
...
+            label = stringResource(id = R.string.site_link),
+            onValueChange = onEditTextValueChange,
+            error = uiState.errorMessage,
+            enabled = false,
```
mikedawson: You need to add fieldsEnabled boolean to the viewModel, and then this enabled property should be controlled by the view moedl

app-android/src/main/java/com/ustadmobile/port/android/view/SiteEnterLinkFragment.kt
```diff
@@ -93,4 +126,81 @@ class SiteEnterLinkFragment : UstadBaseFragment(), SiteEnterLinkView{
...
+                .fillMaxWidth(),
+            elevation = null,
+            colors = ButtonDefaults.buttonColors(
+                backgroundColor = Color.Transparent,
+                contentColor = colorResource(id = R.color.primaryColor),
+            )
+        ) {
+
+            Icon(Icons.Filled.Add, contentDescription = "")
+
```
mikedawson, line 127: Should have a spacer using 

Spacer(Modifier.size(ButtonDefaults.IconSpacing)) 

e.g. as per Google's sample https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/samples/src/main/java/androidx/compose/material/samples/ButtonSamples.kt

app-android/src/main/java/com/ustadmobile/port/android/view/Login2Fragment.kt
ParimSsme: should I add a new variable with mainError name in loginViewModel?

app-android/src/main/java/com/ustadmobile/port/android/view/PersonDetailFragment2.kt
ParimSsme: I will add it's correct value

mikedawson (approved):

PR #418, opened 2022-11-20
Dev mvvm register minor wait for parent

ParimSsme:


core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/RegisterMinorWaitForParentViewModel.kt
```diff
@@ -0,0 +1,10 @@
+package com.ustadmobile.core.viewmodel
+
+data class RegisterMinorWaitForParentUiState(
+
+    val username: String = "",
+
+    val password: String = "",
+
+    val passwordVisible: Boolean = false,
```
mikedawson: Please remove passwordVisible from the UiState class. The ViewModel won't manage this. It should be handled by using state in the composable, like how we handle the password field. This is because there is no logic connected to the password being visible or invisible.

app-android/src/main/java/com/ustadmobile/port/android/view/RegisterMinorWaitForParentFragment.kt
```diff
@@ -60,4 +87,84 @@ class RegisterMinorWaitForParentFragment: UstadBaseFragment(), RegisterMinorWait
...
+
+        Spacer(modifier = Modifier.height(20.dp))
+
+        UstadDetailField(
+            valueText = password,
+            labelText = stringResource(R.string.password),
+            imageId = R.drawable.ic_baseline_vpn_key_24,
+            secondaryActionContent = {
+                IconButton(
+                    onClick = onClickTogglePasswordVisibility,
```
mikedawson: Please update this to use a state within the composable.

mikedawson (commented):
Please remove passwordVisible from the UiState and use state within the compose functions and component (same as what we do for password on the password text field). This is because there is logic connected to the password being visible or not visible, it is only a UI thing.

mikedawson (approved):

PR #422, opened 2022-11-22
Dev mvvm ustad button

ParimSsme:


mikedawson:
I think we should consider the places where our button does not look like the default buttons provided (for example, sometimes we are using left alignment, but the default alignment for a button is supposed to be centered). I think most of the time the right answer will be that we should make our buttons use the default styles.

Google has guidelines on what things should look like here: https://m2.material.io/components/buttons . Let's discuss on call.

mikedawson:
For now - we won't add this because we will try and stick to the default styles for button

PR #423, opened 2022-11-23
Dev site

MunirehHosseinzada:
Adding conversion of Site Details screen


app-android/src/main/java/com/ustadmobile/port/android/view/SiteDetailFragment.kt
```diff
@@ -183,4 +199,49 @@ class SiteDetailFragment: UstadDetailFragment<Site>(), SiteDetailView, Workspace
...
+    onClickLang: (SiteTermsWithLanguage) -> Unit = {}
+){
+
+    Column {
+        UstadDetailField(
+            valueText = uiState.site?.siteName ?: "",
+            labelText = stringResource(R.string.name),
+            imageId = R.drawable.ic_account_balance_black_24dp
+        )
+        UstadDetailField(valueText = stringResource(if(uiState.site?.guestLogin == true){R.string.yes} else {R.string.no}), labelText = "Guest login enabled", imageId = R.drawable.ic_document_preview)
```
mikedawson: These lines are too long. The gray line on the right hand side of the IDE indicates the maximum line length that should be used. E.g. should be split into three lines:

```
UstadDetailField(
            valueText = stringResource(if(uiState.site?.guestLogin == true){R.string.yes} else {R.string.no}), 
            labelText = stringResource(R.string.guest_login_enabled), 
            imageId = R.drawable.ic_document_preview
        )
```

Please make sure to always use stringResource, not just the hardcoded string in English (e.g. labelText = stringResource(..)). Otherwise it will not be translated when the app is used in another language.

app-android/src/main/java/com/ustadmobile/port/android/view/SiteDetailFragment.kt
```diff
@@ -183,4 +199,49 @@ class SiteDetailFragment: UstadDetailFragment<Site>(), SiteDetailView, Workspace
...
+        UstadDetailField(valueText = stringResource(if(uiState.site?.registrationAllowed == true){R.string.yes} else {R.string.no}), labelText = "Registration allowed", imageId = R.drawable.ic_baseline_how_to_reg_24)
+        Text(stringResource(R.string.terms_and_policies), style = Typography.h6)
+
+        uiState.siteTerms.forEach {
+            TextButton(
```
mikedawson: I think we should use the ListItem - it will automatically set the correct styles. See https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#ListItem(androidx.compose.ui.Modifier,kotlin.Function0,kotlin.Function0,kotlin.Boolean,kotlin.Function0,kotlin.Function0,kotlin.Function0) .


e.g. (instead of TextButton):

```
ListItem(
   modifier = Modifier.clickable { onClickLang(lang) },
   text = { Text(lang.stLanguage?.name ?: "")  } 
)
```

app-android/src/main/java/com/ustadmobile/port/android/view/SiteEditFragment.kt
```diff
@@ -128,4 +155,124 @@ class SiteEditFragment: UstadEditFragment<Site>(), SiteEditView {
...
+            error = uiState.siteNameError,
+            enabled = uiState.fieldsEnabled,
+            onValueChange = {
+                onSiteChanged(uiState.site?.shallowCopy{
+                    siteName = it
+                })
+            }
+        )
+
+        CustomSwitch(
```
mikedawson: Please merge dev-mvvm-primary and change this to use UstadSwitchField.

app-android/src/main/java/com/ustadmobile/port/android/view/SiteEditFragment.kt
```diff
@@ -128,4 +155,124 @@ class SiteEditFragment: UstadEditFragment<Site>(), SiteEditView {
...
+                    registrationAllowed = it
+                })
+            }
+        )
+
+        Text(
+            stringResource(R.string.terms_and_policies),
+            style = Typography.h6
+        )
+
```
mikedawson, line 229: On the edit screen we need the option to add an item. The add item is itself a ListItem. It should have an onClick function (e.g. onClickNewTerms). 

Like other lists on Android, the add item option is itself a listItem. It should have an add icon. ListItem has an option for the main icon. 
e.g. something like this
```
ListItem(
  modifier = Modifier.clickable {
     onClickNewTerms()
  }
  text = stringResource(R.string.add_new_terms),
  icon = Icon(
                 imageVector = Icons.Filled.Add,
                 contentDescription = null
)
```
Note: content description is null because the icon is just for decoration, and there is text for adding a new item. It is not like the delete button, where there is no text.

app-react/src/main/kotlin/com/ustadmobile/view/SiteDetailScreen.kt
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            Typography {
+                variant = TypographyVariant.h6
+                + strings[MessageID.terms_and_policies]
+            }
+
+            props.uiState.siteTerms.forEach { item ->
+                ListItem {
+                    var item = item
```
mikedawson, line 62: No need for var item = item, because it is already named as part of the forEach line.

app-react/src/main/kotlin/com/ustadmobile/view/SiteEditScreen.kt
```diff
@@ -0,0 +1,141 @@
...
+                enabled = props.uiState.fieldsEnabled
+                onChange = {
+                    props.onSiteChanged(
+                        props.uiState.site?.shallowCopy {
+                            siteName = it
+                        })
+                }
+            }
+
+            CustomSwitch{
```
mikedawson: Please switch this to using UstadSwitchField

app-react/src/main/kotlin/com/ustadmobile/view/SiteEditScreen.kt
```diff
@@ -0,0 +1,141 @@
...
+                label = strings[MessageID.registration_allowed]
+                checked = props.uiState.site?.registrationAllowed ?: true
+            }
+
+            Typography {
+                variant = TypographyVariant.h6
+                + strings[MessageID.terms_and_policies]
+            }
+
+            List{
```
mikedawson, line 80: Needs to add the add new terms option as per the jetpack compose fragment.

app-react/src/main/kotlin/com/ustadmobile/view/SiteDetailScreen.kt
mikedawson: No need for var item = item, you have already named it as part of the forEach

mikedawson (changes requested):
SiteDetail screens for web and jetpack compose look fine. Thank you. Please see changes required for SiteEdit.

PR #425, opened 2022-11-25
Dev mvvm school detail overview

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/SchoolDetailOverviewFragment.kt
```diff
@@ -170,4 +194,174 @@ class SchoolDetailOverviewFragment: UstadDetailFragment<SchoolWithHolidayCalenda
...
+@Composable
+private fun SchoolDetailOverviewScreen(
+    uiState: SchoolDetailOverviewUiState = SchoolDetailOverviewUiState(),
+    onClickSchoolCode: () -> Unit = {},
+    onClickSchoolPhoneNumber: () -> Unit = {},
+    onClickSms: () -> Unit = {},
+    onClickEmail: () -> Unit = {},
+    onClickClazz: () -> Unit = {},
+) {
+    Column(
```
mikedawson: Because we have a list, please use LazyColumn here. You can then use the item() and items() function as needed

app-android/src/main/java/com/ustadmobile/port/android/view/SchoolDetailOverviewFragment.kt
```diff
@@ -170,4 +194,174 @@ class SchoolDetailOverviewFragment: UstadDetailFragment<SchoolWithHolidayCalenda
...
+
+
+@Composable
+private fun Clazzes(
+    clazzes: Map<Int, ClazzWithListDisplayDetails>,
+    onClickClazz: () -> Unit
+){
+
+    clazzes.forEach { (key, value) ->
+        TextButton(onClick = onClickClazz) {
```
mikedawson: We shouldn't use a button as the base component here. It will change the styles (e.g. alignment, text color, etc). Please use the normal column and row, and then use the modifier.clickable for the onClick function. This will also provide the ripple effect.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/SchoolDetailOverviewViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.ustadmobile.lib.db.entities.ClazzWithListDisplayDetails
+import com.ustadmobile.lib.db.entities.SchoolWithHolidayCalendar
+
+data class SchoolDetailOverviewUiState(
+
+    var entity: SchoolWithHolidayCalendar? = null,
+
+    val schoolCodeVisible: Boolean = false,
+
+    val clazzes: Map<Int, ClazzWithListDisplayDetails> = emptyMap()
```
mikedawson: DataSource.Factory should be represented as a list instead of a map.

app-android/src/main/java/com/ustadmobile/port/android/view/SchoolDetailOverviewFragment.kt
mikedawson: Also we should use the Jetpack compose official ListItem to manage the styles (even though it is experimental): https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#ListItem(androidx.compose.ui.Modifier,kotlin.Function0,kotlin.Function0,kotlin.Boolean,kotlin.Function0,kotlin.Function0,kotlin.Function0)

mikedawson: e.g. 

```
ListItem(
   modifier = Modifier.clickable {
      onClickClazz(clazz)
    },
    text = { Text(clazz.clazzName ?: "") },
    secondaryText = { Text(clazz.clazzDesc ?: "") }  
)
```

mikedawson (approved):

mikedawson (approved):

PR #427, opened 2022-11-26
Dev mvvm primary

MunirehHosseinzada:
Site Detail page, web conversion


mikedawson:
This request is targeting the main primary and dev-mvvm-primary. This will only be done when dev-mvvm-primary is completely finished. I think the branch here must have been selected by mistake.

PR #431, opened 2022-11-30
Dev mvvm content entry edit

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/composable/CourseBlockEdit.kt
```diff
@@ -0,0 +1,252 @@
...
+
+@Composable
+fun CourseBlockEdit(
+    uiState: CourseBlockEditUiState,
+    onCourseBlockChange: (CourseBlock?) -> Unit = {}
+){
+
+    Column{
+
+        FirstRow(
```
mikedawson: Let's not use extra functions for each row... just put them in one function. e.g. delete the functions FirstRow, SecondRow etc and move the content inside CourseBlockEdit.

app-android/src/main/java/com/ustadmobile/port/android/view/composable/CourseBlockEdit.kt
```diff
@@ -0,0 +1,252 @@
...
+    }
+}
+
+@Composable
+private fun FourthRow(
+    uiState: CourseBlockEditUiState,
+    onCourseBlockChange: (CourseBlock?) -> Unit
+){
+    Row {
+        UstadDateEditTextField(
```
mikedawson: There will be one new component UstadDateTimeEditTextFields - you can switch to using that. Then one composable will include both date and time.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -459,4 +479,233 @@ class ContentEntryEdit2Fragment(
 
     }
 
+}
+
+@Composable
+private fun ContentEntryEditScreen(
+    uiState: ContentEntryEditUiState = ContentEntryEditUiState(),
+    courseBlockEditUiState: CourseBlockEditUiState = CourseBlockEditUiState(),
```
mikedawson: CourseBlockEditUiState is already part of ContentEntryEditUiState - we don't need to have it as an additional parameter. You can use uiState.courseBlockEditState.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryEdit2Fragment.kt
```diff
@@ -459,4 +479,233 @@ class ContentEntryEdit2Fragment(
...
+                onContentChanged(uiState.entity?.shallowCopy {
+                        licenseType = it.value
+                    }
+                )
+            }
+        )
+
+        Spacer(modifier = Modifier.height(15.dp))
+
+        if (uiState.containerStorageOptionVisible){
```
mikedawson, line 610: Add selectedContainerStorageDir as a variable on UiState

Can use:

UstadExposedDropDownMenuField(
        value = uiState.selectedContainerStorageDir,
        label = stringResource(R.string..),
        options = uiState.containerStorageDirs,
        onOptionSelected = {
            onSelectContainerStorageDir(it)
        }
        itemText = { it.name ?: "" },
        enabled = uiState.fieldsEnabled,
    )

app-android/src/main/java/com/ustadmobile/port/android/view/composable/UstadCourseBlockEdit.kt
```diff
@@ -0,0 +1,134 @@
...
+                    onCourseBlockChange(uiState.courseBlock?.shallowCopy{
+                        cbCompletionCriteria = it.value
+                    })
+                },
+                enabled = uiState.fieldsEnabled,
+            )
+
+            Spacer(modifier = Modifier.width(15.dp))
+
+            if (uiState.minScoreVisible){
```
mikedawson, line 58: This will crash:

The value of the text will be something like "5 points". If you try "5 points".toInt() - it will crash because the string is not valid for parsing into an int.

The "points" string or percentage etc needs to added as an adornment (like a trailing icon), not in the value of the string itself. I have added a property called suffix to UstadTextEditField on React and Jetpack compose that will do this. You can use this for the points string and the percentage string. I have also included examples in the previews.

We also need to set the keyboard type to numeric so that users will see only the numbers on the keyboard, not all the letters. I have added KeyboardOptions as a parameter on UstadTextEditField, which is then passed to the textfield we are using. This is also included in the example. There is an example also for setting the inputType on React (this will have an effect if the user opens the web version using a mobile browser).

Finally, we need to filter out anything the user might try to type that is not a number to be completely sure that we won't have invalid input. On Android it is still possible to connect a bluetooth keyboard, or copy/paste something invalid. We cannot rely only on showing the numerical keyboard.

For that we need to adjust the onChange function to use something like this:

```
onChange = { newString ->
   uiState.entity.numericalProperty = newString.filter { it.isDigit() }.toIntOrNull() ?: 0
}
```

mikedawson (approved):

PR #432, opened 2022-12-01
Dev mvvm content entry detail overview

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +376,438 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+            modifier = Modifier.size(110.dp),
+            contentScale = ContentScale.Crop
+        )
+
+        Row(
+            verticalAlignment = Alignment.CenterVertically
+        ) {
+            if (uiState.scoreProgressVisible){
+                LinearProgressIndicator(
+                    progress = 0.6F,
```
mikedawson: What to display here must be based on the scoreProgress, should not use hard-coded demo constants here! The progress bar should only show when there is progress (e.g. > 0). Also the checkbox should only show if the user has passed/completed. The cross should be shown if the user failed. If the user has not passed, completed, or failed, then there is no icon at the end of the checkbox.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +376,438 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+                )
+            }
+        }
+
+        Row(
+            verticalAlignment = Alignment.CenterVertically
+        ){
+
+            if (uiState.fileSizeVisible){
+                Text(text = uiState.contentEntry?.container?.fileSize.toString())
```
mikedawson: This needs to use UMFileUtil.formatFileSize . The fileSize on the property is in bytes. formatFileSize will show in KB, MB, etc according to the size.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +376,438 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+                    contentDescription = "",
+                    modifier = Modifier.size(18.dp)
+                )
+
+                Text(uiState.scoreProgress?.progress.toString())
+            }
+
+            Spacer(modifier = Modifier.width(16.dp))
+
+            Text("(" +
```
mikedawson: This must only show when there is a score to show.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +376,438 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+                ) {
+                    Text(stringResource(R.string.download).uppercase(),
+                        color = contentColorFor(
+                            colorResource(id = R.color.primaryColor))
+                    )
+                }
+            }
+        }
+
+        item {
```
mikedawson: It would be better to use padding instead of extra items to manage spacing. This is especially important when there is an item that the user can tap. The tap area needs padding to make it big enough for the user. This can only happen when using padding.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +376,438 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+                onClickDelete = onClickDelete,
+                onClickManageDownload = onClickManageDownload
+            )
+        }
+
+        item {
+            Spacer(modifier = Modifier.height(10.dp))
+        }
+
+        item {
```
mikedawson: This should be removed. Not sure where this from the design. Also we should not be using hardcoded strings instead of stringResource.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +376,438 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+
+        if (uiState.translationVisibile){
+            item {
+                Text(text = stringResource(id = R.string.also_available_in))
+            }
+
+            items(
+                uiState.availableTranslations.size
+            ){
+                ListItem(
```
mikedawson: The available translations is a list... but we should show this as buttons in rows as we are doing using XML. I think this can be done using https://google.github.io/accompanist/flowlayout/ . We are only showing the name of the language, so one language per row is taking up too much space. 

e.g. we should have:

English, Persian, Arabic, French
German

Instead of:

English
Persian
Arabic
French
German

We can't just use a column because there might be a lot of translations, we need something that will make a new line when required.

app-android/src/main/java/com/ustadmobile/port/android/view/RegisterAgeRedirectFragment.kt
```diff
@@ -4,30 +4,14 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import androidx.compose.foundation.layout.*
```
mikedawson: This file is not related to the screen. It should not be changed in this pull request.

mikedawson:
Because this pull request contains unrelated files (SiteTerms, RegisterAgeRedirect, etc), we need a new pull request and a new branch for this

PR #438, opened 2022-12-06
Dev mvvm language edit

MunirehHosseinzada:


app-android/src/main/java/com/ustadmobile/port/android/view/LanguageEditFragment.kt
```diff
@@ -72,4 +86,68 @@ class LanguageEditFragment: UstadEditFragment<Language>(), LanguageEditView {
...
+){
+    Column (
+        modifier = Modifier
+            .padding(8.dp)
+            .verticalScroll(rememberScrollState()),
+    horizontalAlignment = Alignment.Start
+    ){
+
+        UstadTextEditField(
+            value = stringResource(id = R.string.name),
```
mikedawson: Value for all fields should come from the entity on the UI state class, should not use the string constant. This appears to be done correctly on the react version, but not here on the jetpack compose version.

e.g. should be:

value - uiState.language.name
label = stringResource(R.string.name)

mikedawson (changes requested):
Value fields needs to use the language entity itself on UIState. Values on Android now are using the label constants.

PR #440, opened 2022-12-08
Dev mvvm holiday calendar detail

MunirehHosseinzada:


core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/HolidayCalendarDetailViewModel.kt
```diff
@@ -0,0 +1,8 @@
+package com.ustadmobile.core.viewmodel
+
+import com.ustadmobile.lib.db.entities.HolidayCalendar
+
+data class HolidayCalendarDetailUIState(
+    val holidayCalendar: HolidayCalendar? = null,
+    val calendarList: List<HolidayCalendar>? = emptyList()
```
mikedawson: The HolidayCalendar is basically a name linked to a list of Holidays. Each Holiday has a holiday start time and end time (e.g. Eid, summer vacation, winter vacation, etc).

For example there could be a HolidayCalendar for the cold area schedule schools and warm area schedule schools.

e.g. 
```
val holidayCalendar: HolidayCalendar? = null
val holidayList: List<Holiday> = emptyList()
```

Please change the list to being Holiday instead of HolidayCalendar.

Lists do not need to be nullable because they can be empty

app-android/src/main/java/com/ustadmobile/port/android/view/HolidayCalendarDetailFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+            .verticalScroll(rememberScrollState()),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+
+        UstadDetailField(
+            valueText = uiState.holidayCalendar?.umCalendarName ?: "",
+            labelText = stringResource(id = R.string.name)
+        )
+
+        uiState.calendarList?.forEach {
```
mikedawson: Please name the variable in these lists so they have a meaningful name. These are slightly longer more complex functions

e.g.

```
uiState.holidayList.forEach { holiday ->
   
}
```

app-android/src/main/java/com/ustadmobile/port/android/view/HolidayCalendarDetailFragment.kt
```diff
@@ -0,0 +1,70 @@
...
+            ListItem (
+                modifier = Modifier
+                    .clickable {
+                        onItemClick(it)
+                    },
+
+                text = { Text(it.umCalendarName ?: "") }
```
mikedawson: secondaryText should be the holiday start and end date. You can get the formatted dates using the rememberFormattedDate function

e.g. 

```
val holidayStart = rememberFormattedDate(holiday.holStartTime, "UTC")
val holidayEnd = rememberFormattedDate(holiday.holEndTime, "UTC)

... 

secondaryText = { Text("$holidayStart - $holidayEnd") }
```

The second parameter is the timezone. The timezone for holidays is always UTC.

app-react/src/main/kotlin/com/ustadmobile/view/HolidayCalendarDetailScreen.kt
```diff
@@ -0,0 +1,72 @@
...
+                props.uiState.calendarList?.forEach { item ->
+                    ListItem {
+
+                        disablePadding = true
+
+                        ListItemButton {
+                            onClick = {
+                                props.onItemClick(item)
+                            }
+                            ListItemText {
```
mikedawson, line 50: Please add secondary text for the holiday start and end time as per the Jetpack compose version. On React the function is useFormattedDate (instead of rememberFormattedDate ).

mikedawson (changes requested):
Please update as per comments.

app-react/src/main/kotlin/com/ustadmobile/view/HolidayCalendarDetailScreen.kt
mikedawson: useMemo is close - but should be done using useFormattedDate. useFormattedDate will format based on the language setting of the user (which might be different to the browser default).

app-android/src/main/java/com/ustadmobile/port/android/view/HolidayCalendarDetailFragment.kt
```diff
@@ -0,0 +1,82 @@
...
+import com.ustadmobile.port.android.view.composable.UstadDetailField
+
+@Composable
+fun HolidayCalendarDetailScreen(
+    uiState: HolidayCalendarDetailUIState,
+    onItemClick: (Holiday) -> Unit = {}
+){
+    Column(
+        modifier = Modifier
+            .padding(8.dp)
```
mikedawson: Only vertical padding should be set here. Horizontal padding needs to be set by the components themselves. If there is a ripple effect, horizontal padding on the container would stop the ripple from reaching the edge. 

Therefor we need to set the horizontal padding on each item, not using the container.

Will look into standardizing how we are handling padding later this evening.

app-android/src/main/java/com/ustadmobile/port/android/view/HolidayCalendarDetailFragment.kt
```diff
@@ -0,0 +1,82 @@
...
+            .verticalScroll(rememberScrollState()),
+        horizontalAlignment = Alignment.CenterHorizontally
```
mikedawson: Why is this using center horizontally? I don't think we need to change the default (e.g. left aligned for English, right aligned for dari/pashto/arabic etc).

app-react/src/main/kotlin/com/ustadmobile/view/HolidayCalendarDetailScreen.kt
```diff
@@ -0,0 +1,87 @@
...
+                        val holidayEnd = useMemo(item.holEndTime) {
+                            Date(item.holEndTime).toLocaleDateString()
+                        }
+
+                        ListItemButton {
+                            onClick = {
+                                props.onItemClick(item)
+                            }
+                            ListItemText {
+                                + (item.holName ?: "")
```
mikedawson: You need to use the primary and secondary properties for ListItemText instead of putting the text in directly. ListItemSecondaryAction will appear in the top right corner. That is for secondary action buttons (e.g. download buttons, option menus etc) as per the MaterialDesign guidelines ( https://m2.material.io/components/lists#anatomy ) . 

E.g. it should look like this:

```
ListItemText {
    primary = ReactNode(item.holName ?: "")
    secondary = ReactNode("$holidayStart - $holidayEnd")
}
```

mikedawson (changes requested):
This is structurally correct, but please update as per comments to handle lists etc correctly. Thanks.

mikedawson:
This pull request had unrelated files added by accident. This screen has already been merged via new pull request.

PR #443, opened 2022-12-11
Dev mvvm clazz edit

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,376 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
...
+                },
+                text = { Text(stringResource(id = R.string.add_block)) },
+            )
+        }
+
+        items (
+            items = uiState.courseBlockList
+        ) { courseBlock ->
+
+            val CONTENT_ENTRY_TYPE_ICON_MAP = mapOf(
```
mikedawson: This map is a constant - so let's move it outside of the composable function (for performance). We don't need to make a new copy of this map every time the function runs, and then repeat that for each block.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,376 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
...
+    onClickTerminology: () -> Unit = {},
+    onClickHideBlockPopupMenu: (CourseBlockWithEntity?) -> Unit = {},
+    onClickIndentBlockPopupMenu: (CourseBlockWithEntity?) -> Unit = {},
+    onClickUnIndentBlockPopupMenu: (CourseBlockWithEntity?) -> Unit = {},
+    onClickDeleteBlockPopupMenu: (CourseBlockWithEntity?) -> Unit = {},
+) {
+    LazyColumn(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
```
mikedawson: Do we need the Spacer items when we have the padding? Doesn't padding already solve the problem for us?

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,376 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
...
+                ContentEntry.TYPE_AUDIO to R.drawable.ic_audiotrack_24px
+            )
+            val image = if(courseBlock.cbType == CourseBlock.BLOCK_CONTENT_TYPE)
+                courseBlock.entry?.contentTypeFlag
+            else
+                courseBlock.cbType
+
+            val courseBlockEditAlpha: Float = if (courseBlock.cbHidden) 0.5f else 1f
+
+            ListItem(
```
mikedawson: To implement the indent: the start padding of the ListItem should be 8dp * the indent level of the block.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,376 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
...
+                icon = {
+                    Icon(
+                        imageVector = Icons.Filled.Add,
+                        contentDescription = "",
+                    )
+                }
+            )
+        }
+
+        items(
```
mikedawson, line 475: better to use:
```
items(
 items = uiState.clazzSchedules
) { schedule ->
   
}
```

We don't need to access items by using an array index. I think this is clearer than using it as an index.

Also: the Schedules are part of a List, so they should use ListItem, not UstadDetailField.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,393 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
     }
 
 
+}
+
+private val CONTENT_ENTRY_TYPE_ICON_MAP = mapOf(
```
mikedawson: This also needs to include the icon maps for CourseBlock as per CourseBlock.kt:

        const val BLOCK_MODULE_TYPE = 100

        const val BLOCK_TEXT_TYPE = 102

        const val BLOCK_ASSIGNMENT_TYPE = 103

        const val BLOCK_CONTENT_TYPE = 104

        const val BLOCK_DISCUSSION_TYPE = 105

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,393 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
...
+        }
+
+        item {
+            Text(text = stringResource(id = R.string.course_blocks))
+        }
+
+        item {
+            ListItem(
+                modifier = Modifier.clickable {
+                    onClickAddCourseBlock()
```
mikedawson: Please add the plus icon for the add course block item.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,393 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
...
+            uiState.clazzSchedules
+        ){ schedule ->
+
+            val fromTimeFormatted = rememberFormattedTime(timeInMs = schedule.sceduleStartTime.toInt())
+            val toTimeFormatted = rememberFormattedTime(timeInMs = schedule.scheduleEndTime.toInt())
+            val text = "${messageIdResource(id = schedule.scheduleFrequency)} " +
+                    " ${messageIdResource(schedule.scheduleDay)} " +
+                    " $fromTimeFormatted - $toTimeFormatted "
+
+            ListItem(
```
mikedawson, line 485: Please add an invisible icon (e.g. alpha = 0)  to the schedule ListItem so that the alignment matches with the add button.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzEditFragment.kt
```diff
@@ -284,4 +317,393 @@ class ClazzEditFragment() : UstadEditFragment<ClazzWithHolidayCalendarAndSchoolA
...
+                courseBlock.entry?.contentTypeFlag
+            else
+                courseBlock.cbType
+
+            val courseBlockEditAlpha: Float = if (courseBlock.cbHidden) 0.5F else 1F
+            val startPadding = (courseBlock.cbIndentLevel * 8).dp
+
+            ListItem(
+                modifier = Modifier
+                    .padding(start = startPadding)
```
mikedawson: There needs to be an onClickCourseBlock event handler.

mikedawson (approved):

PR #445, opened 2022-12-14
Dev mvvm scoped grant edit

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/ScopedGrantEditFragment.kt
```diff
@@ -87,4 +100,55 @@ class ScopedGrantEditFragment: UstadEditFragment<ScopedGrant>(), ScopedGrantEdit
...
+    onChangedBitmask: (BitmaskFlag?) -> Unit = {},
+) {
+
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
+    ) {
+        uiState.bitmaskList.forEach { bitmask ->
+
+            ListItem(
```
mikedawson, line 118: Using the switch inside a listitem like this is OK and is a valid use case as per the documentation

ListItem here needs to use modifier.toggleable. Using the current version, you can only tap on the switch itself. The user should also be able to tap anywhere on the text to toggle the switch.

e.g.

```
ListItem(
  modifier = modifier.toggleable( 
     role = Role.Switch,
     value = bitmask.enabled,
     onValueChange = {
         onBitmaskChanged(bitmask.copy(enabled = it))
     }
  ),
  text = { Text(messageIdResource(...) },
  trailing =  {
        Switch(
            //Note: do not use the event handler on Switch. It is handled by the toggleable on container
            checked = bitmask.enabled,
        )
    }
)
```

app-react/src/main/kotlin/com/ustadmobile/view/ScopedGrantEditScreen.kt
```diff
@@ -0,0 +1,66 @@
...
+val ScopedGrantEditScreenComponent2 = FC<ScopedGrantEditScreenProps> { props ->
+
+    val strings = useStringsXml()
+
+    Container {
+
+        List{
+            props.uiState.bitmaskList
+                .forEach { bitmask ->
+                ListItem{
```
mikedawson, line 27: Please use ListItemButton and it's onClick function so that clicking anywhere on the item will toggle the switch.

mikedawson (changes requested):
The structure here is good, some relatively minor changes need to make sure that clicking anywhere on the item will toggle it (e.g. not only clicking on the Switch, but clicking on the text etc).

mikedawson:
This has already been merged. This pull request is closed because it includes unrelated files.

PR #448, opened 2022-12-15
Dev mvvm report filter edit

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/ReportFilterEditFragment.kt
```diff
@@ -272,4 +294,173 @@ class ReportFilterEditFragment : UstadEditFragment<ReportFilter>(), ReportFilter
...
+                        Icon(
+                            imageVector = Icons.Filled.Delete,
+                            contentDescription = stringResource(id = R.string.delete),
+                        )
+                    }
+                }
+            )
+        }
+
+        item {
```
mikedawson, line 432: The add filter button should be above the filter items, not below them.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportFilterEditFragment.kt
```diff
@@ -272,4 +294,173 @@ class ReportFilterEditFragment : UstadEditFragment<ReportFilter>(), ReportFilter
...
+                )
+            }
+        }
+
+        items(
+            items = uiState.uidAndLabelList
+        ){ uidAndLabel ->
+            ListItem(
+                text = {
+                       TextButton(onClick = { onClickEditFilter(uidAndLabel)  }) {
```
mikedawson, line 417: Should not use TextButton. The ListItem is not clickable.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportFilterEditFragment.kt
```diff
@@ -272,4 +294,173 @@ class ReportFilterEditFragment : UstadEditFragment<ReportFilter>(), ReportFilter
...
+                        })
+                    }
+                )
+            }
+        }
+
+        items(
+            items = uiState.uidAndLabelList
+        ){ uidAndLabel ->
+            ListItem(
```
mikedawson, line 415: Please add an empty icon (e.g. UstadBlankIcon) so that the text aligns with the add button (which has the plus icon).

app-android/src/main/java/com/ustadmobile/port/android/view/ReportFilterEditFragment.kt
```diff
@@ -272,4 +294,173 @@ class ReportFilterEditFragment : UstadEditFragment<ReportFilter>(), ReportFilter
...
+        item {
+            ListItem(
+                modifier = Modifier.clickable {
+                    onClickNewItemFilter()
+                },
+                text = { Text(uiState.createNewFilter) },
+                icon = {
+                    Icon(
+                        imageVector = Icons.Filled.Add,
+                        contentDescription = stringResource(id = R.string.add),
```
mikedawson, line 441: contentDescription here should be null or blank. There is already text for this item (e.g. it is not just an icon on it's own). The icon is just for appearance. 

We need the contentDescription on delete because there is no text for that clickable item (there is only the icon for the trash). For adding, we have text.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportFilterEditFragment.kt
```diff
@@ -272,4 +294,173 @@ class ReportFilterEditFragment : UstadEditFragment<ReportFilter>(), ReportFilter
...
+                    onOptionSelected = {
+                        onReportFilterChanged(uiState.reportFilter?.shallowCopy{
+                            reportFilterDropDownValue = it.value
+                        })
+                    },
+                )
+            }
+        }
+
+        item {
```
mikedawson, line 363: Please add a reportFilterValueVisible field on UiState. Then this should only visible when reportFilterValueVisible is true.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportFilterEditFragment.kt
```diff
@@ -272,4 +294,173 @@ class ReportFilterEditFragment : UstadEditFragment<ReportFilter>(), ReportFilter
...
+                keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Number),
+                onValueChange = {
+                    onReportFilterChanged(uiState.reportFilter?.shallowCopy{
+                        reportFilterValue = it
+                    })
+                }
+            )
+        }
+
+        item {
```
mikedawson, line 378: Please add a reportFilterBetweenValueVisible field on UiState. Then this should only visible when reportFilterBetweenValueVisible is true.

app-android/src/main/java/com/ustadmobile/port/android/view/ReportFilterEditFragment.kt
```diff
@@ -272,4 +294,173 @@ class ReportFilterEditFragment : UstadEditFragment<ReportFilter>(), ReportFilter
...
+                    onValueChange = {
+                        onReportFilterChanged(uiState.reportFilter?.shallowCopy{
+                            reportFilterValueBetweenY = it
+                        })
+                    }
+                )
+            }
+        }
+
+        items(
```
mikedawson, line 412: Please add reportFilterUidAndLabelListVisible boolean to UiState. This should only be visible of that is true.

app-react/src/main/kotlin/com/ustadmobile/view/ReportFilterEditScreen.kt
```diff
@@ -0,0 +1,201 @@
...
+                            Delete {}
+                            onClick = {
+                                props.onClickRemoveFilter(uidAndLabel)
+                            }
+                        }
+                    }
+                }
+            }
+
+            ListItem {
```
mikedawson, line 185: The ListItem for add a new filter should be above. It should also be inside the list.

app-react/src/main/kotlin/com/ustadmobile/view/ReportFilterEditScreen.kt
```diff
@@ -0,0 +1,201 @@
...
+                props.uiState.uidAndLabelList.forEach { uidAndLabel ->
+                    ListItem{
+                        ListItemText {
+                            primary = ReactNode(uidAndLabel.labelName ?: "")
+                            onClick = {
+                                props.onClickEditFilter(uidAndLabel)
+                            }
+                        }
+                        secondaryAction = IconButton.create {
+                            Delete {}
```
mikedawson, line 176: Please add ariaLabel property so that this button is accessible (same as contentDescription on Jetpack)
e.g.
```
IconButton.create {
    ariaLabel = strings[MessageID.delete]
    onClick = {  
       ...
    }
    Delete { } 
}
```

mikedawson (changes requested):
Thank you, please see comments for changes needed.

PR #449, opened 2022-12-15
Dev mvvm clazz log edit

ParimSsme:


mikedawson:
Unfortunately this also has unrelated files, including CourseTerminology. Please make a new pull request with only the related files.

PR #450, opened 2022-12-18
Scoped Grant Detail Page

MunirehHosseinzada:


app-android/src/main/java/com/ustadmobile/port/android/view/ScopedGrantDetailFragment.kt
```diff
@@ -109,4 +122,48 @@ class ScopedGrantDetailFragment: UstadDetailFragment<ScopedGrantWithName>(
...
+){
+
+    Column (
+        modifier = Modifier
+            .fillMaxSize()
+    ){
+        uiState.bitmaskList.forEach { bitmask ->
+            ListItem(
+                modifier = Modifier.clickable {
+                    onItemClick(bitmask)
```
mikedawson: The bitmask flag itself on detail mod should not be clickable. It won't go anywhere.

app-android/src/main/java/com/ustadmobile/port/android/view/ScopedGrantDetailFragment.kt
```diff
@@ -109,4 +122,48 @@ class ScopedGrantDetailFragment: UstadDetailFragment<ScopedGrantWithName>(
...
+                },
+                text = {
+                    Text(messageIdResource(id = bitmask.messageId))
+                }
+            )
```
mikedawson, line 152: Please add a trailing icon to show a check or a cross (e.g. check if it is enabled, cross if not enabled).

app-react/src/main/kotlin/com/ustadmobile/view/ScopedGrantDetailScreen.kt
```diff
@@ -0,0 +1,59 @@
...
+
+val ScopedGrantDetailComponent2 = FC<ScopedGrantDetailProps> { props ->  
+
+    Container{
+
+        val strings = useStringsXml()
+
+        props.uiState.bitmaskList.forEach { bitmask ->
+            ListItem{
+                ListItemButton{
```
mikedawson: As per the Jetpack version - this is not clickable, so please don't use ListItemButton, just use ListItem. Also as per the Jetpack version, please add a trailing (e.g. secondaryAction) component to show a check or a cross depending on whether or not it is enabled.

mikedawson (changes requested):
Thanks, please see the changes requested as per comments.

mikedawson (approved):

PR #451, opened 2022-12-20
Content Entry Import Link

MunirehHosseinzada:


app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryImportLinkFragment.kt
```diff
@@ -83,4 +103,67 @@ class ContentEntryImportLinkFragment : UstadBaseFragment(), ContentEntryImportLi
     }
 
 
+}
+
+@Composable
+fun ContentEntryImportLinkScreen(
+    uiState: ContentEntryImportLinkUiState,
+    onClickNext: () -> Unit = {},
+    onUrlChange: (Url?) -> Unit = {}
```
mikedawson: This needs to use a string type. The user might type an invalid URL (the final URL might be invalid, or it might be something is invalid as typed and then the user changes it). 

This code could crash in that case. For example let's say the user starts typing a link, and there is just one letter. One letter on its own is not a valid URL. The onValueChange function will run, calling Url("h") could crash. Or maybe as per the preview we get something other than what we expected (e.g. http://localhost/h). 

We need to keep this as a String.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/ContentEntryImportLinkViewModel.kt
```diff
@@ -0,0 +1,9 @@
+package com.ustadmobile.core.viewmodel
+
+import io.ktor.http.*
+
+data class ContentEntryImportLinkUiState(
+    var url: Url? = null,
```
mikedawson: This needs to be a string as per previous comment

app-react/src/main/kotlin/com/ustadmobile/view/ContentEntryImportLinkScreen.kt
```diff
@@ -0,0 +1,72 @@
...
+import mui.material.styles.TypographyVariant
+import mui.system.responsive
+import mui.system.sx
+import react.FC
+import react.Props
+
+external interface ContentEntryImportLinkProps: Props {
+    var uiState: ContentEntryImportLinkUiState
+    var onClickNext: () -> Unit
+    var onUrlChange: (Url?) -> Unit
```
mikedawson: Needs to change from Url type to String type.

mikedawson (changes requested):
Thanks. Can you please update as per feedback?

PR #453, opened 2022-12-27
Dev mvvm coursediscussion

varunasingh:


app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetail2Fragment.kt
```diff
@@ -0,0 +1,355 @@
...
+    ) {
+        itemsIndexed(posts) { _, item ->
+            PostDetailItem(item)
+        }
+    }
+
+}
+
+@Composable
+private fun AddPostButton(onClick: () -> Unit) {
```
mikedawson, line 212: The floating action button should not be here. The floating action button has to be in the activity, not the fragment. The floating action button will be controlled by the viewmodel. As per recommended design, the ViewModel will have a flow that will emit an AppState (which can include the floating action button text, icon, visibility, event handler, title for the app, etc).

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetail2Fragment.kt
```diff
@@ -0,0 +1,355 @@
...
+
+@Composable
+private fun CourseDiscussionDetailScreen(
+    uiState: CourseDiscussionDetailUiState = CourseDiscussionDetailUiState(),
+    onClickAddPost: () -> Unit = {},
+    onClickPost: (DiscussionPostWithDetails) -> Unit = {},
+    onClickDeletePost: (DiscussionPostWithDetails) -> Unit = {},
+){
+
+    Column(
```
mikedawson, line 278: This needs to use LazyColumn. The posts themselves will be loaded lazily using PagedList (the replacement for DataSourceFactory). You can use a plain list for now (I still need to add PagedList to door - which will be straightforward).

You can then use something like this:


```
LazyColumn  {
    item {
       //description etc here
    }

    items(
        items = uiState.posts,
        key = { it.postUid }
     ) { post ->
         //item content goes here
     }
}
```

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetail2Fragment.kt
```diff
@@ -0,0 +1,355 @@
...
+        modifier = Modifier
+            .fillMaxSize()
+            .verticalScroll(rememberScrollState())
+    ) {
+        Text(
```
mikedawson, line 283: I think the title will be in the title bar (on the activity). It shouldn't be here in the fragment.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetail2Fragment.kt
```diff
@@ -0,0 +1,355 @@
...
+            modifier = Modifier.padding(8.dp)
+        )
+        Spacer(modifier = Modifier.height(10.dp))
+        Text(stringResource(R.string.posts),
+            style = Typography.h4,
+            modifier = Modifier.padding(8.dp))
+
+
+        Box(modifier = Modifier.weight(1f)){
+            PostsDetail(uiState.posts, onClickPost)
```
mikedawson, line 301: I think it would be better to use the Jetpack compose ListItem instead of creating something additional here. ListItem will automatically apply the correct padding, spacing, typography, click ripple effect etc for a material design list. It has been used in other screens. I also notice that right now the onClickPost is not really connected anywhere, so clicking won't do anything. Something like this:

```
ListItem(
   modifier = modifier.clickable {
      onClickPost(post)
   },
   icon = {
      //Avatar icon here
   },
   text = { 
      Text(postNames)
   },
   secondaryText = {
      Column {
         Text(discussionPostTitle)
         Text(postLatestMessage)
      }
   },
   singleLineSecondaryText = false,
   trailing = {
      Column {
          Text(date)
          Text(stringResource(R.string.num_replies, post.numReplies))
       }
   }
)
```

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetail2Fragment.kt
```diff
@@ -0,0 +1,355 @@
...
+        }
+
+        Column(
```
mikedawson, line 304: Floating action button should not be inside the fragment, this is managed by the activity.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetail2Fragment.kt
```diff
@@ -0,0 +1,355 @@
+package com.ustadmobile.port.android.view
```
mikedawson, line 1: For the Jetpack compose UI, we are putting that inside the existing fragments. Please move the jetpack compose code to the existing fragment.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionEdit2Fragment.kt
```diff
@@ -0,0 +1,356 @@
+package com.ustadmobile.port.android.view
```
mikedawson, line 1: The posts themselves should not be in the edit. The edit is only allowing the course leader to change the title, description, and don't show before date.

There is a CourseBlockEdit function that can be used (it is used by ContentEntryEdit). 

The entity being edited should actually be the CourseBlock. You can't have a Discussion without the courseblock, so we might as well use those fields for title / description.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionEdit2Fragment.kt
```diff
@@ -0,0 +1,356 @@
...
+    ) {
+        itemsIndexed(posts) { _, item ->
+            PostDetailItem(item)
+        }
+    }
+
+}
+
+@Composable
+private fun PostDetailItem(post: DiscussionPostWithDetails){
```
mikedawson, line 185: Identical copy/pasted code (duplicated from CourseDiscussionDetail)? DRY!

app-react/src/main/kotlin/com/ustadmobile/view/CourseDiscussionBlockEditScreen.kt
```diff
@@ -0,0 +1,194 @@
...
+
+
+
+        Typography {
+            variant = TypographyVariant.h6
+            +strings[MessageID.posts
+            ]
+        }
+
+        List {
```
mikedawson, line 97: Posts shouldn't be in the edit section.

app-react/src/main/kotlin/com/ustadmobile/view/CourseDiscussionBlockEditScreen.kt
```diff
@@ -0,0 +1,194 @@
...
+                        ListItemIcon {
+                            UstadBlankIcon { }
+                        }
+
+                        onClick = {
+                            props.onClickPost(item)
+                        }
+
+
+                        val dateFormatted = useMemo(dependencies = arrayOf(item.postLatestMessageTimestamp)) {
```
mikedawson, line 136: This is the correct use of dependencies for the useMemo function. Better to use the function called "useFormattedDate" (which I added). This will use the locale set by the user (which works correctly when the user chooses a different locale to the browser default e.g. browser set to English, user selects Tajik). 

useFormattedDate requires the timezone. You can use the system default timezone via the new kotlinx datetime 

```
import kotlinx.datetime.TimeZone

TimeZone.currentSystemDefault().id
```

app-react/src/main/kotlin/com/ustadmobile/view/CourseDiscussionDetailScreen.kt
```diff
@@ -0,0 +1,182 @@
...
+
+                        onClick = {
+                            props.onClickPost(item)
+                        }
+
+                        val dateFormatted = useMemo(dependencies = arrayOf(item.postLatestMessageTimestamp)) {
+                            Date(item.postLatestMessageTimestamp ?: 0L).toLocaleDateString()
+                        }
+
+                        UstadMessageField{
```
mikedawson, line 124: We shouldn't use this inside a ListItem. I think we should be using ListItem as is. The formatting and alignment goes wrong because there is a blank icon in the ListItem itself, and then an icon in the content.

app-react/src/main/kotlin/com/ustadmobile/view/CourseDiscussionDetailScreen.kt
```diff
@@ -0,0 +1,182 @@
...
+                ListItem {
+                    disablePadding = true
+//                    secondaryAction = IconButton.create {
+//                        onClick = {
+//                            props.onDeleteClick(item)
+//                        }
+//                        Delete {}
+//                    }
+
+                    ListItemButton {
```
mikedawson, line 111: The replies are not clickable, so we should not be using ListItemButton.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetail2Fragment.kt
```diff
@@ -0,0 +1,355 @@
...
+
+
+
+@Composable
+private fun PostsDetail(
+    posts: List<DiscussionPostWithDetails> = emptyList(),
+    onClickPost: (DiscussionPostWithDetails) -> Unit = {}
+){
+
+    LazyColumn(
```
mikedawson, line 200: Nested LazyColumn won't work (its the same as nesting a recyclerview). It needs to be the top level container. The good news is that within a LazyColumn you can easily use item / items, no need for concatadapter etc.

mikedawson (changes requested):
Thanks. Please see comments. Might be best to switch to making one pull request per screen so it can be merged and reviewed faster.

mikedawson:
Closing this pull request as it has been superseded by per-screen pull requests.

PR #455, opened 2022-12-27
End 2 end test- Courses creation, enrolment, attendance

poojaustad:
Courses creation, enrolment, attendance


test-end-to-end/android-maestro/common/accepting_course_join_request.yaml
```diff
@@ -0,0 +1,13 @@
...
+- tapOn:
+      id: "com.toughra.ustadmobile:id/item_clazzlist_clazz_cl"
+- tapOn: "MEMBERS"
+- tapOn:
+    id: "com.toughra.ustadmobile:id/accept_icon"
+- extendedWaitUntil:
+    notVisible:
+      id: "com.toughra.ustadmobile:id/accept_icon"
+    timeout: 10000
+- takeScreenshot: "results/StudentGotAdded"
```
mikedawson: I think any screenshot that we take using the common yaml files needs to be a variable. If this common file is used by multiple different flows, then the screenshot will just get replaced.

test-end-to-end/android-maestro/common/add_new_people.yaml
```diff
@@ -0,0 +1,20 @@
+appId: com.toughra.ustadmobile
```
mikedawson, line 1: I think the password should be a variable, rather than hardcoded.

test-end-to-end/android-maestro/common/add_new_student_inside_course.yaml
```diff
@@ -0,0 +1,30 @@
...
+- tapOn: "Add a new person"
+- tapOn: "First names"
+- inputText: ${STUDENTFIRSTNAME}
+- tapOn: "Last name"
+- inputText: ${STUDENTLASTNAME}
+- tapOn: "Gender"
+- tapOn: ${GENDER}
+- tapOn: "SAVE"
+- tapOn: "SAVE"
+- swipe:
```
mikedawson, line 27: What is this swipe for? Can you add a comment to the source? Are we sure those exact numbers will always work?

test-end-to-end/android-maestro/common/member_login.yaml
```diff
@@ -0,0 +1,15 @@
+appId: com.toughra.ustadmobile
```
mikedawson: I think there isn't really a different login flow for admin or anyone else. The name/structure is a little confusing.

Let's have one flow for login that accepts a variable for the username and password. I think we should call it LOGIN_USER or USER etc - PERSONNAME is confusing (makes me think we are talking about the person's actual name, not the username). 

I think it would be better to have flows per screen e.g.

get_started.yaml: just do hideKeyboard, tapOn getStarted
select_site.yaml - enter the endpoint and click next. Can accept a variable for the endpoint.
Login - accepts a variable for the username and password.

Then to start one would use three flows:

```
runFlow get_started
runFlow select_site
   # Set endpoint vars
runFlow login
    # set login username and password var
```

test-end-to-end/android-maestro/common/course_with_members.yaml
```diff
@@ -0,0 +1,87 @@
+appId: com.toughra.ustadmobile
```
mikedawson: Let's call this file create_course_with_members - because creating the course is what it does.

test-end-to-end/android-maestro/common/edit_recorded_attendance.yaml
```diff
@@ -0,0 +1,19 @@
+appId: com.toughra.ustadmobile
+---
+- tapOn: "Courses"
+- tapOn:
+      id: "com.toughra.ustadmobile:id/item_clazzlist_clazz_cl"
+- tapOn: "ATTENDANCE"
+- tapOn:
+    point: "58%,81%"
```
mikedawson: Why this exact point? Are we sure these coordinates will always work? Can you please add a comment to the source here to explain why we are using these literal values?

mikedawson (changes requested):
Thanks! Looks generally good, please see comments.

test-end-to-end/android-maestro/common/accepting_course_join_request.yaml
poojaustad: Ok Mike, got it. Thanks.

test-end-to-end/android-maestro/common/add_new_student_inside_course.yaml
poojaustad: There is a known issue with the maestro. Scroll functionality is not working for the emulator that we are using in Jenkins. So we are using the swipe command for the test. It's working as expected in this test.

test-end-to-end/android-maestro/common/course_with_members.yaml
poojaustad: Sure Mike

test-end-to-end/android-maestro/common/member_login.yaml
poojaustad: Got it Mike, I'll make changes ,Thank you

test-end-to-end/android-maestro/common/edit_recorded_attendance.yaml
poojaustad: Hi Mike, Yes, using these literal values is not the right practice. ID is available for this element. I have changed the point command to id. Thank you.

test-end-to-end/android-maestro/common/get_started.yaml
```diff
@@ -0,0 +1,6 @@
+appId: com.toughra.ustadmobile
```
mikedawson, line 1: Lets call this clear_state_and_get_started

test-end-to-end/android-maestro/common/add_new_people.yaml
poojaustad: Done Mike, Thanks

test-end-to-end/android-maestro/common/add_new_student_inside_course.yaml
mikedawson: I see the comment added for this, thanks. Got it.

mikedawson (approved):

mikedawson:
Merged, thank you @poojaustad !

PR #458, opened 2022-12-28
dev-mvvm-contentEntryDetailOverview2

ParimSsme:
…pose


app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +377,466 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+            )
+        }
+
+        if (uiState.contentEntryButtons?.showDownloadButton == true){
+            item {
+                Button(
+                    onClick = onClickDownload,
+                    modifier = Modifier
+                        .fillMaxWidth(),
+                    colors = ButtonDefaults.buttonColors(
```
mikedawson: The button color background color will be the primary color of the theme. There is no need to specify the colors here, because you will already get the same by leaving the defaults. You also don't need to specify the color on the text within the button. 

You only need to specify the color when you want to use a non default value (e.g. secondary color).

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +377,466 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+                        color = contentColorFor(
+                            colorResource(id = R.color.primaryColor))
+                    )
+                }
+            }
+        }
+
+        if (uiState.contentEntryButtons?.showOpenButton == true){
+            item {
+                Button(
```
mikedawson, line 430: Same as above - no need to specify colors, can use defaults.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +377,466 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+    }
+}
+
+@Composable
+fun ContentDetails(
+    uiState: ContentEntryDetailOverviewUiState = ContentEntryDetailOverviewUiState(),
+){
+    Row {
+
+        Box(
```
mikedawson: Please rename LeftColumn and RightColumn to ContentDetailLeftColumn and ContentDetailRightColumn - otherwise it is not really clear what they are. I think then we don't need ContentDetails as it's own function. 

I think the extra Box is not needed. The Modifier.weight can be used within the left/right column functions directly.

Right now contentdetails is really just a row and a spacer, so I think that does not need it's own function.

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +377,466 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+        BadgedBox(badge = { Badge {
+            Icon(
+                painter = painterResource(id = R.drawable.ic_content_complete),
+                contentDescription = "Favorite"
+            )
+        } }) {
+            if (uiState.scoreProgressVisible){
+                LinearProgressIndicator(
+                    progress = (uiState.scoreProgress?.progress ?: 0)
+                        .toFloat()/10.0F,
```
mikedawson: Why divide by 10, shouldn't that be 100?

app-android/src/main/java/com/ustadmobile/port/android/view/ContentEntryDetailOverviewFragment.kt
```diff
@@ -346,4 +377,466 @@ class ContentEntryDetailOverviewFragment: UstadDetailFragment<ContentEntryWithMo
...
+){
+    Column{
+        Image(painter = painterResource(id = R.drawable.book_24px),
+            contentDescription = "",
+            modifier = Modifier.size(110.dp),
+            contentScale = ContentScale.Crop
+        )
+
+        BadgedBox(badge = { Badge {
+            Icon(
```
mikedawson: What icon to show needs to be set based on the status. Maybe we should use the built-in icons inside the badge. Color of the badge can be set based on the status (e.g. pass = green, fail = red). 

Seems like it needs some more spacing between the columns (badge overlaps the text on the right).

Screenshot from 2022-12-28 13-53-14

app-react/src/main/kotlin/com/ustadmobile/view/ContentEntryDetailOverviewScreen.kt
```diff
@@ -0,0 +1,483 @@
...
+
+    var onClickOpen: () -> Unit
+
+    var onClickMarkComplete: () -> Unit
+
+    var onClickDelete: () -> Unit
+
+    var onClickManageDownload: () -> Unit
+
+    var onClickTranslation: () -> Unit
```
mikedawson: This needs to take the ContentEntryRelatedEntryJoinWithLanguage as a parameter for the event - otherwise the event handler won't know what language was clicked.

app-react/src/main/kotlin/com/ustadmobile/view/ContentEntryDetailOverviewScreen.kt
```diff
@@ -0,0 +1,483 @@
+package com.ustadmobile.view
```
mikedawson, line 1: Please add this to useustadscreens so that when running it will appear in the list.

app-react/src/main/kotlin/com/ustadmobile/view/ContentEntryDetailOverviewScreen.kt
```diff
@@ -0,0 +1,483 @@
...
+            }
+
+            Divider { orientation = Orientation.horizontal }
+
+            if (props.uiState.translationVisibile){
+                Typography{
+                    + strings[MessageID.also_available_in]
+                }
+
+                Grid {
```
mikedawson, line 108: Grid is a good way to show this. Generally usage of grid looks correct to me.

app-react/src/main/kotlin/com/ustadmobile/view/ContentEntryDetailOverviewScreen.kt
```diff
@@ -0,0 +1,483 @@
...
+                            md = 1
+
+                            Box {
+                                sx {
+                                    padding = 8.px
+                                    textAlign = TextAlign.center
+                                }
+
+                                Button {
+                                    onClick = { props.onClickTranslation }
```
mikedawson: Needs variant = ButtonVariant.text to show the correct button style (e.g. textbutton), should not have background color.

mikedawson (changes requested):
Thanks, please see comments.

mikedawson (approved):
All looks good, thank you!

PR #462, opened 2022-12-28
CourseTerminologyEdit

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/CourseTerminologyEditFragment.kt
```diff
@@ -93,4 +104,81 @@ class CourseTerminologyEditFragment: UstadEditFragment<CourseTerminology>(), Cou
...
+    onCtTitleChanged: (String?) -> Unit = {}
+) {
+    LazyColumn(
+        modifier = Modifier
+            .fillMaxSize()
+            .padding(16.dp)
+    )  {
+
+        item{
+            ListItem(
```
mikedawson: Because we are using textfields, we should not put them inside a ListItem. They are not really normal ListItems (e.g. they don't have a title, subheader, icon, etc).

app-android/src/main/java/com/ustadmobile/port/android/view/CourseTerminologyEditFragment.kt
```diff
@@ -93,4 +104,81 @@ class CourseTerminologyEditFragment: UstadEditFragment<CourseTerminology>(), Cou
...
+                text = {
+                    UstadTextEditField(
+                        value = uiState.entity?.ctTitle ?: "",
+                        label = stringResource(id = R.string.name),
```
mikedawson: This should be messageIdResource - the messageids that are on common (e.g. terminologyTerm.messageId shared between MUI and Android) have different int values.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseTerminologyEditFragment.kt
```diff
@@ -93,4 +104,81 @@ class CourseTerminologyEditFragment: UstadEditFragment<CourseTerminology>(), Cou
...
+
+@Composable
+@Preview
+fun CourseTerminologyEditScreenPreview() {
+    val uiState = CourseTerminologyEditUiState(
+        terminologyTermList = listOf(
+            TerminologyEntry(
+                id = "1",
+                term = "First",
+                messageId = R.string.message
```
mikedawson: This should use MessageID e.g.
messageID = MessageID.teacher , MessageID.student, etc.

app-react/src/main/kotlin/com/ustadmobile/view/CourseTerminologyEditScreen.kt
```diff
@@ -0,0 +1,104 @@
...
+                error = props.uiState.titleError
+                enabled = props.uiState.fieldsEnabled
+                onChange = {
+                    props.onCtTitleChanged(it)
+                }
+            }
+
+            + strings[MessageID.your_words_for]
+
+            List {
```
mikedawson: This should not use List/ListItem as above - we don't have a List with header and subheaders as per the materialdesign spec. We just have a column / stack of textfields.

app-react/src/main/kotlin/com/ustadmobile/view/CourseTerminologyEditScreen.kt
```diff
@@ -0,0 +1,104 @@
...
+
+val CourseTerminologyEditScreenPreview = FC<Props> {
+
+    val uiStateVal by useState {
+        CourseTerminologyEditUiState(
+            terminologyTermList = listOf(
+                TerminologyEntry(
+                    id = "1",
+                    term = "First",
+                    messageId = MessageID.message
```
mikedawson: Let's use MessageID.teacher, MessageID.student, MessageID.add_a_teacher for the preview. These are the terms that the admin can actually change.

mikedawson (changes requested):
Thanks - overall approach looks good. Some changes to make as per feedback.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseTerminologyEditFragment.kt
ParimSsme: ok, what is this terminologyTerm.messageId?

mikedawson (approved):

PR #465, opened 2022-12-28
ReportEdit

ParimSsme:


app-react/src/main/kotlin/com/ustadmobile/view/ReportEditScreen.kt
```diff
@@ -0,0 +1,300 @@
...
+                enabled = props.uiState.fieldsEnabled
+                onChange = {
+                    props.onReportChanged(
+                        props.uiState.report?.shallowCopyReportWithSeriesWithFilters{
+                            reportDateRangeSelection = it?.value ?: 0
+                        })
+                }
+            }
+
+            List{
```
mikedawson: List should only be used to contain ListItem. It should be used for the filters and maybe the add series button, but shouldn't contain textfields.

app-react/src/main/kotlin/com/ustadmobile/view/ReportEditScreen.kt
```diff
@@ -0,0 +1,300 @@
...
+                        onClickRemoveSeries = props.onClickRemoveSeries
+                        onClickNewFilter = props.onClickNewFilter
+                        onClickDeleteReportFilter = props.onClickDeleteReportFilter
+                        onChangedReportSeries = props.onChangedReportSeries
+                    }
+                }
+            }
+
+            ListItem {
+                onClick = { props.onClickNewSeries() }
```
mikedawson: ListItem that contains onClick in React must use ListItemButton

app-react/src/main/kotlin/com/ustadmobile/view/ReportEditScreen.kt
```diff
@@ -0,0 +1,300 @@
...
+            }
+        }
+    }
+}
+
+private val ReportSeriesListItem = FC<ReportEditScreenProps> { props ->
+
+    val strings = useStringsXml()
+
+    ListItem {
```
mikedawson: Shouldn't use ListItem here.

mikedawson (changes requested):
Please don't use ListItem to contain textfields. Please make sure that when ListItem is used (for the list of filters) that the ListItems are all together inside one List, and use ListItemButton for clicakble items.

PR #467, opened 2022-12-28
Holiday Calendar Edit 2

MunirehHosseinzada:


app-react/src/main/kotlin/com/ustadmobile/view/HolidayCalendarEditScreen.kt
```diff
@@ -0,0 +1,123 @@
...
+                enabled = props.uiState.fieldsEnabled
+                onChange = {
+                    props.onHolidayCalendarChange(
+                        props.uiState.holidayCalendar?.shallowCopy {
+                            umCalendarName = it
+                        })
+                }
+            }
+
+            ListItem {
```
mikedawson, line 54: This ListItem also needs to go inside the List, otherwise the spacing will be wrong.

app-android/src/main/java/com/ustadmobile/port/android/view/HolidayCalendarEditFragment.kt
```diff
@@ -145,4 +164,93 @@ class HolidayCalendarEditFragment() : UstadEditFragment<HolidayCalendar>(), Holi
...
+            text = { Text(stringResource(id = R.string.add_a_holiday)) }
+        )
+
+        uiState.holidayList?.forEach { holiday ->
+
+            val holidayStart = rememberFormattedDate(timeInMillis = holiday.holStartTime, timeZoneId = "UTC")
+            val holidayEnd = rememberFormattedDate(timeInMillis = holiday.holEndTime, timeZoneId = "UTC")
+
+            ListItem(
+                modifier = Modifier
```
mikedawson, line 215: The holiday itself does not have an icon, but you need to use a blank icon so that the spacing will be correct and the text will line up. The text for "add a holiday" and "Eid" should align, 

spacer

The way to do this is to add a blank icon so that the text will start at the same position.

e.g. add this to the listitem.

```
ListItemIcon {
   UstadBlankIcon { } 
}
```

app-react/src/main/kotlin/com/ustadmobile/view/HolidayCalendarEditScreen.kt
```diff
@@ -0,0 +1,123 @@
...
+
+                    onClick = {
+                        props.onAddItemClick()
+                    }
+                }
+            }
+
+            List{
+                props.uiState.holidayList?.forEach { holiday ->
+                    ListItem {
```
mikedawson, line 74: The holiday itself does not have an icon, but you need to use a blank icon so that the spacing will be correct and the text will line up. The text for "add a holiday" and "Eid" should align, 

spacer

The way to do this is to add a blank icon so that the text will start at the same position.

e.g. add this to the listitem.

```
ListItemIcon {
   UstadBlankIcon { } 
}
```

mikedawson, line 75: The holiday itself does not have an icon, but you need to use a blank icon so that the spacing will be correct and the text will line up. The text for "add a holiday" and "Eid" should align, 

spacer

The way to do this is to add a blank icon so that the text will start at the same position.

e.g. add this to the listitem.

```
ListItemIcon {
   UstadBlankIcon { } 
}
```

app-android/src/main/java/com/ustadmobile/port/android/view/HolidayCalendarEditFragment.kt
mikedawson, line 217: The holiday itself does not have an icon, but we need to make the text align (e.g. "Add a holiday" and "Eid" should align and start at the same position.

spacer

You can do this by adding a spacer like this:

```
ListItem(
   icon = {
      Spacer(modifier = Modifier.width(24.dp))
   }
}
```

mikedawson (changes requested):
Please change the minor remaining feedback items.

mikedawson (approved):
Looks good, thank you!

PR #475, opened 2023-01-04
Course Discussion MVVM

varunasingh:


core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/CourseDiscussionDetailViewModel.kt
```diff
@@ -0,0 +1,10 @@
+package com.ustadmobile.core.viewmodel
+
+import com.ustadmobile.lib.db.entities.CourseDiscussion
+import com.ustadmobile.lib.db.entities.DiscussionPostWithDetails
+
+data class CourseDiscussionDetailUiState(
+    val courseDiscussion: CourseDiscussion? = null,
```
mikedawson: I think as per call we don't need CourseDiscussion entity. Title, description, etc are on the courseblock.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetailFragment.kt
```diff
@@ -145,6 +185,124 @@ class CourseDiscussionDetailFragment: UstadDetailFragment<CourseDiscussion>(),
...
+        }
+
+
+        items(
+            items = uiState.posts,
+            key = { post -> post.discussionPostUid }
+        ){ post ->
+
+            val context = LocalContext.current
+            val datePosted = remember { DateFormat.getDateFormat(context)
```
mikedawson: Please use rememberFormattedDate. This doesn't have a key set, so we haven't told compose what would invalidate it. There needs to be an explicit invalidation key, e.g. the timestamp. This is done by the rememberformatteddate function.

app-android/src/main/java/com/ustadmobile/port/android/view/CourseDiscussionDetailFragment.kt
```diff
@@ -145,6 +185,124 @@ class CourseDiscussionDetailFragment: UstadDetailFragment<CourseDiscussion>(),
...
+                        imageVector = Icons.Filled.Person,
+                        contentDescription = null
+                    )
+                },
+                text = {
+                    Text(post.discussionPostMessage ?: "" )
+                },
+                secondaryText = {
+                    Column {
+                        Text(post.discussionPostTitle?: "")
```
mikedawson: Shouldn't the title be in the primary text, then the latest message is in the secondary text? I don't think we need the post message itself.

app-android/src/main/java/com/ustadmobile/port/android/view/composable/UstadMessageField.kt
```diff
@@ -0,0 +1,201 @@
...
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.toughra.ustadmobile.R
+
+/**
+ * Main field layout for message fields in DetailViews
+ */
+@Composable
+fun UstadMessageField(
```
mikedawson: This is not used, so it should be removed.

app-react/src/main/kotlin/com/ustadmobile/view/CourseDiscussionDetailScreen.kt
```diff
@@ -0,0 +1,173 @@
...
+
+                        onClick = {
+                            props.onClickPost(item)
+                        }
+
+                        val dateFormatted = useMemo(dependencies = arrayOf(item.postLatestMessageTimestamp)) {
+                            Date(item.postLatestMessageTimestamp ?: 0L).toLocaleDateString()
+                        }
+
+                        UstadMessageField{
```
mikedawson: As per discussion, we should be using the standard ListItem and its formatting.

app-react/src/main/kotlin/com/ustadmobile/mui/components/UstadMessageField.kt
```diff
@@ -0,0 +1,266 @@
...
+import mui.system.Stack
+import mui.system.StackDirection
+import mui.system.responsive
+import mui.system.sx
+import react.FC
+import react.Props
+import react.ReactNode
+import react.create
+
+external interface UstadMessageFieldProps: Props {
```
mikedawson: This should be removed as we should be using the standard ListItem

mikedawson (changes requested):
Thanks, please see the comments. Please remove the use of UstadMessageField and use the standard ListItem on React/MUI. Please remove  the (unused) UstadMessageField on Jetpack compose.

mikedawson:
Already updated/ merged / updated in dev-mvvm-primary

PR #482, opened 2023-01-12
Dev mvvm clazz log edit attendance1

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/ClazzLogEditAttendanceFragment.kt
```diff
@@ -325,4 +362,232 @@ class ClazzLogEditAttendanceFragment: UstadEditFragment<ClazzLog>(), ClazzLogEdi
...
+        },
+        icon = {
+            Icon(
+                painter = painterResource(id = R.drawable.ic_person_black_24dp),
+                contentDescription = ""
+            )
+        },
+        trailing = {
+
+            var selectedButtonId: Int? = null
```
mikedawson: This kind of variable usage won't work in compose/react functions. It will get recreated every time the function runs. It is not a state and it cannot be used to coordinate anything between function calls. 

Please remove it and use only the function parameters to set the UI in this function. 

e.g. update can set the checked item using a map lookup:

//Buttons Map is a constant and should be declared outside the function - no need to create it every time the function runs
val BUTTONS_MAP = mapOf(status to buttonId)

onUpdate you can use something like

buttonGroup.check(BUTTONS_ID_MAP[clazzLog.attendanceStatus])
//else - need to handle situation when nothing is checked.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzLogEditAttendanceFragment.kt
```diff
@@ -325,4 +362,232 @@ class ClazzLogEditAttendanceFragment: UstadEditFragment<ClazzLog>(), ClazzLogEdi
...
+        }
+    }
+}
+
+@OptIn(ExperimentalPagerApi::class)
+@SuppressLint("CoroutineCreationDuringComposition")
+@Composable
+private fun PagerView(
+    list: List<ClazzLog>,
+    timeZone: String
```
mikedawson: This needs an onChangeClazzLog function to tell the ViewModel to load / display the data for the date when the teacher changes the day.

app-react/src/main/kotlin/com/ustadmobile/view/ClazzLogEditAttendanceScreen.kt
```diff
@@ -0,0 +1,279 @@
...
+
+    var onClickNextClazzLog: () -> Unit
+
+    var clazzLog: List<ClazzLog>
+
+}
+
+private val PagerView = FC<PagerViewProps> { props ->
+
+    val dateTime = useFormattedDateAndTime(
```
mikedawson, line 157: This is hardcoded to  use ClazzLog 0. It needs to use a state variable inside. Clicking needs to trigger the onChangeClazzLog function (so it uses the same function as Android).

app-react/src/main/kotlin/com/ustadmobile/view/ClazzLogEditAttendanceScreen.kt
```diff
@@ -0,0 +1,279 @@
...
+    var clazzLog: ClazzLogAttendanceRecordWithPerson
+
+    var onClazzLogAttendanceChanged: (ClazzLogAttendanceRecordWithPerson) -> Unit
+
+    var fieldsEnabled: Boolean
+}
+
+private val ClazzLogItemView = FC<ClazzLogItemViewProps> { props ->
+
+    val iconList = listOf(
```
mikedawson: This can be made as a map once outside the function. We don't need to create them every time the function runs. It is not going to be a significant performance issue, but where possible we should do things once instead of repeating them every time a function runs.

mikedawson (changes requested):
Thank you! Please see the comments.

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzLogEditAttendanceFragment.kt
```diff
@@ -325,4 +362,239 @@ class ClazzLogEditAttendanceFragment: UstadEditFragment<ClazzLog>(), ClazzLogEdi
 
     }
 
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+private fun ClazzLogEditAttendanceScreen(
+    uiState: ClazzLogEditAttendanceUiState = ClazzLogEditAttendanceUiState(),
+    onClickMarkAllPresent: (Int) -> Unit = {},
```
mikedawson: We only need one function: onClickMarkAll. The Int argument is the status (attended or absent). Please change this to having only one onClickMarkAll function.

You can use parameter names for function types where the type is not enough information to know the meaning: e.g.

```
onClickMarkAll: (status: Int) -> Unit = {}
```

mikedawson (approved):

PR #489, opened 2023-01-18
ClazzMemberListScreen

ParimSsme:


core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/ClazzMemberListViewModel.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.ustadmobile.core.db.dao.ClazzEnrolmentDaoCommon
+import com.ustadmobile.core.generated.locale.MessageID
+import com.ustadmobile.core.util.MessageIdOption2
+import com.ustadmobile.core.util.SortOrderOption
+import com.ustadmobile.lib.db.entities.PersonWithClazzEnrolmentDetails
+
+data class ClazzMemberListUiState(
+
+    val studentList: List<PersonWithClazzEnrolmentDetails> = emptyList(),
```
mikedawson, line 12: We need to use the course terminology e.g. if the terminology uses the word "course participant" instead of student, we should show that.

Please merge the latest dev-mvvm-primary and add CourseTerminologyStrings to the UiState.

For any Message ID mentioned in TerminologyKeys , get it from TerminologyStrings first e.g.

uiState.terminologyStrings[MessageID.add_a_student] ?: stringResource(R.string.add_a_student)

app-android/src/main/java/com/ustadmobile/port/android/view/ClazzMemberListFragment.kt
```diff
@@ -265,4 +276,296 @@ class ClazzMemberListFragment() : UstadListViewFragment<PersonWithClazzEnrolment
...
+}
+
+ @OptIn(ExperimentalMaterialApi::class)
+ @Composable
+ fun StudentListItem(
+     person: PersonWithClazzEnrolmentDetails,
+     onClick: (PersonWithClazzEnrolmentDetails) -> Unit,
+ ){
+
+     val statusColor = if ((person.attendance/100) >=
```
mikedawson: Please create a top level function called colorForAttendanceStatus(attendance) because we are using this in other screens. You can put this in a separate file in com.ustadmobile.port.android.view.util - we can call it ColorForAttendanceStatus.

Please make a similar function for React.

mikedawson (changes requested):
The screen looks good: two changes needed:

1. Please use the TerminologyStrings for the headers and "add a teacher" / "add a student" button as per the comment.
2. Please create a function that will return the attendance color in a common location for react/compose because this will be used on other screens.

Thanks!

PR #497, opened 2023-01-24
UstadAddListItem

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/composable/UstadAddListItem.kt
```diff
@@ -0,0 +1,55 @@
...
+fun UstadAddListItem(
+    text: String,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    icon: ImageVector = Icons.Filled.Add,
+    onClickAdd: (() -> Unit) = {  }
+){
+    ListItem(
+        modifier = modifier
+            .defaultItemPadding()
```
mikedawson: We shouldn't add default item padding here... ListItem already has padding

app-react/src/main/kotlin/com/ustadmobile/mui/components/UstadAddListItem.kt
```diff
@@ -0,0 +1,52 @@
...
+import react.ReactNode
+import react.create
+
+external interface UstadAddListItemProps: Props {
+
+    var text: String
+
+    var enabled: Boolean?
+
+    var icon: ReactNode
```
mikedawson: We want the add icon to be the default, but we can't set default values on an external interface. 

You can achieve the same effect by making this nullable:

e.g. var icon: ReactNode?

then in ListItemIcon use:

+ (props.icon ?: Add.create() )

mikedawson (approved):

PR #515, opened 2023-02-26
Dev authenticator

mikedawson:
Adds support for users to grant permission to an external app (e.g. for single sign-on etc).


mikedawson:
Passes all end-to-end tests and the main build tests.

PR #528, opened 2023-03-15
Number text field

ParimSsme:


app-android/src/main/java/com/ustadmobile/port/android/view/composable/UstadNumberTextEditField.kt
```diff
@@ -0,0 +1,100 @@
...
+                }
+            }else {
+                null
+            },
+            keyboardOptions = KeyboardOptions(
+                keyboardType = KeyboardType.Number,
+                imeAction = ImeAction.Done
+            ),
+            keyboardActions = KeyboardActions(
+                onDone = {
```
mikedawson: This will only run if the user clicks the done checkbox on the keyboard, I suggest removing min/max. Let this be handled by the viewmodel

app-android/src/main/java/com/ustadmobile/port/android/view/composable/UstadNumberTextEditField.kt
```diff
@@ -0,0 +1,100 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.input.ImeAction
+import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun UstadNumberTextEditField(
+    value: String,
```
mikedawson: The value of the number field won't be a string, it would be a number.

PR #600, opened 2023-10-19
Dev test mvvm primary cypress checklist course block

poojaustad:
Checklist 3 - 3 tests are ready


test-end-to-end/webapp-cypress/cypress/e2e/003_006_user_able_to_expand _and_collapse_modules.cy.js
```diff
@@ -0,0 +1,50 @@
...
+        cy.contains("button","Save").click()
+
+     //Testing Module Collapse icon working or not
+        cy.get('svg[data-testid="KeyboardArrowUpIcon"]').eq(0).click()   // collapse module 1
+        cy.contains('text 1').should('not.exist');
+        cy.get('svg[data-testid="KeyboardArrowUpIcon"]').click({force: true})
+        cy.contains("text 2").should('not.exist')
+
+     //Testing Module Expand icon working or not
+        cy.get('svg[data-testid="KeyboardArrowDownIcon"]').eq(0).click()
```
mikedawson: What is the .eq(0) for?

test-end-to-end/webapp-cypress/cypress/e2e/005_003_all_user_add_post_and_reply.cy.js
```diff
@@ -0,0 +1,105 @@
...
+          cy.contains("Courses").should('be.visible')
+          cy.contains('005_003').click()
+
+         // Add reply to the post board
+
+          cy.get('[data-testid="ForumIcon"]').click()
+          cy.contains('Topic 1').click()
+          cy.get('[data-placeholder="Add a reply"]').type('Reply 2 to the topic')
+          cy.contains('button','Post').click()
+          cy.get('.MuiBox-root.css-z7mtfw').eq(1).should('be.visible')
```
mikedawson: Framework -generated css classes (e.g. css-z7mtfw etc) must not be used in selectors under any circumstances. They could change when libraries are updated or if any part of the style of an element is changed, which would then break the test.

If needed we can add an attribute or css class manually in the code.

test-end-to-end/webapp-cypress/cypress/e2e/005_004_user_add_internal_links_to_post.cy.js
```diff
@@ -0,0 +1,91 @@
...
+
+     cy.get('[data-testid="ForumIcon"]').click()
+     cy.contains('Topic 1').click()
+     cy.get('[data-placeholder="Add a reply"]').type('Internal link');
+     cy.get('[data-placeholder="Add a reply"]').type('{selectall}')
+     cy.get('.ql-link').click()
+     cy.get('[data-video="Embed URL"]').type('http://localhost:8087/umapp/#/ContentEntries')
+     cy.get('[class=ql-action]').click()
+     cy.contains('button','Post').click()
+     cy.get('.MuiBox-root.css-z7mtfw').eq(1).should('be.visible')
```
mikedawson: As above: uses auto-generated css selector.

test-end-to-end/webapp-cypress/cypress/e2e/005_005_user_add_external_links_to_post.cy.js
```diff
@@ -0,0 +1,91 @@
...
+
+     cy.get('[data-testid="ForumIcon"]').click()
+     cy.contains('Topic 1').click()
+     cy.get('[data-placeholder="Add a reply"]').type('External link');
+     cy.get('[data-placeholder="Add a reply"]').type('{selectall}')
+     cy.get('.ql-link').click()
+     cy.get('[data-video="Embed URL"]').type('https://github.com/UstadMobile/UstadMobile/blob/primary/test-end-to-end/README.md')
+     cy.get('[class=ql-action]').click()
+     cy.contains('button','Post').click()
+     cy.get('.MuiBox-root.css-z7mtfw').eq(1).should('be.visible')
```
mikedawson: As above re. autogenerated css. This assertion doesn't fulfill any real purpose because what "muibox-root.css-z7mtfw " is not clear, so why must it be visible?

test-end-to-end/webapp-cypress/cypress/e2e/005_004_user_add_internal_links_to_post.cy.js
```diff
@@ -0,0 +1,91 @@
...
+     cy.go('back')
+     cy.go('back')
+      })
+
+     it('Student able to open the internal link in the reply', () => {
+
+         // Student Login
+
+          cy.ustadClearDbAndLogin('studentA','test1234')
```
mikedawson: Where student logs in afterwards, there should be an assertion that the reply made by the teacher should be visible.

mikedawson (changes requested):
This is mostly there for two checklists, which is good. 

Before merging:
1) All use of auto-generated css classes needs to be removed.
2) When a second user is logging in to post a reply, we should have assertions to confirm that the content posted by the previous user (e.g. post, reply, etc) is visible.

mikedawson (approved):

PR #781, opened 2024-01-02
Dev mvvm primary chat

ParimSsme:


app-react/src/jsMain/kotlin/com/ustadmobile/mui/components/Sidebar.kt
```diff
@@ -31,6 +33,7 @@ data class RootScreen(
 val ROOT_SCREENS = listOf(
     RootScreen(ClazzListViewModel.DEST_NAME_HOME, MR.strings.courses, School),
```
mikedawson: When importing material icons on React, we should use something like:

```
import mui.icons.material.School as SchoolIcon
```

Because if Android Studio automatically converts this to import mui.icons.material.* it will make Android Studio run extremely slow or crash.

I think this was missed before on this file. Can you please update the import and use of Icons on this file accordingly?

app-react/src/jsMain/kotlin/com/ustadmobile/view/message/conversationlist/ConversationListScreen.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    ListItem {
+        ListItemButton{
+            onClick = {
+                props.message?.also { props.onListItemClick(it) }
+            }
+
+            ListItemIcon {
+                UstadPersonAvatar {
+                    personUid = props.message?.otherPerson?.personUid ?: 0
```
mikedawson: UstadPersonAvatar actually doesn't need personUid anymore. You need to pass the personName and pictureUri.

app-react/src/jsMain/kotlin/com/ustadmobile/view/message/messagelist/MessageListScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+                backgroundColor = theme.palette.primary.light
+                color = theme.palette.primary.contrastText
+                padding = 10.px
+                margin = 5.px
+                borderTopLeftRadius = 48.px
+                borderTopRightRadius = 48.px
+                borderBottomLeftRadius = if (isFromMe) 48.px else 0.px
+                borderBottomRightRadius = if (isFromMe) 0.px else 48.px
+            }
+            + UstadLinkify.create {
```
mikedawson: In react we have two ways of adding components: passing a ReactNode object and invoking the component function. For example on label we have something like:

```
TextField {
    label = ReactNode("label text")
}
```

If we don't want to use plain text, then a React Node object can be created using .create e.g. 
```
TextField {
   label = Typography.create {
       ...
   }
}
```

Inside another node ("ChildrenBuilder" scope), then :

```
+ ComponentType.create { 
   propName = "value"
} 
```
Is the same as 
```
ComponentType {
    propName = "value"
} 
```

So when adding a component to another component that supports children, we would normally not use + ComponentType.create { }, use use ComponentType { .. }

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/message/conversationlist/ConversationListViewModel.kt
```diff
@@ -1,13 +1,101 @@
...
+                    text = systemImpl.getString(MR.strings.message),
+                    icon = FabUiState.FabIcon.ADD,
+                    onClick = this@ConversationListViewModel::onClickAdd
+                )
+            )
+        }
+    }
+
+    override fun onUpdateSearchResult(searchText: String) {
+        TODO("Not yet implemented")
```
mikedawson: This should be implemented as per other screens so that the user can search by the name of the person they are looking for. You can use the LIKE clause in the SQL query. 

e.g. 
```
WHERE searchQuery = '' OR ((Person.firstNames | ' ' | Person.lastName) LIKE searchQuery)
```

mikedawson:
Closing as chat has been merged via another pull request

PR #839, opened 2024-02-13
Added content move test

poojaustad:
User able to move a content to new folder


test-end-to-end/webapp-cypress/cypress/e2e/001_003_move_content.cy.js
```diff
@@ -0,0 +1,51 @@
...
+})
+
+it('Admin user move content to folder', () => {
+ // Admin user login
+  cy.ustadClearDbAndLogin('admin','testpass',{timeout:8000})
+ // Add H5p File
+  cy.ustadAddContentToLibrary('../test-files/content/H5p_Content.h5p','Content_001')
+  cy.contains('Content_001').click()
+  cy.contains('OPEN').click({force: true})
+  cy.get('#appbar_title').should('be.visible').invoke('text').should('eq', 'Content_001');
```
mikedawson: Wouldn't ```cy.contains("#appbar_title", "Content_001").should("be.visible")``` work here?

test-end-to-end/webapp-cypress/cypress/e2e/001_003_move_content.cy.js
```diff
@@ -0,0 +1,51 @@
...
+    cy.ustadStartTestServer()
+})
+
+it('Admin user move content to folder', () => {
+ // Admin user login
+  cy.ustadClearDbAndLogin('admin','testpass',{timeout:8000})
+ // Add H5p File
+  cy.ustadAddContentToLibrary('../test-files/content/H5p_Content.h5p','Content_001')
+  cy.contains('Content_001').click()
+  cy.contains('OPEN').click({force: true})
```
mikedawson: Why do we need force here? We shouldn't need it and normally shouldn't use it. If the button is disabled by the UI or not visible... it should wait for the timeout.

test-end-to-end/webapp-cypress/cypress/support/commands.js
```diff
@@ -59,7 +59,15 @@ Cypress.Commands.add('ustadAddContentToLibrary',(contentPath,contentName) => {
   cy.get('input[id="content_title"]').click()
   cy.get('input[id="content_title"]').clear().type(contentName,{timeout: 2000})
   cy.get('#actionBarButton').click()
+})
 
+// Add a Folder to Library
+Cypress.Commands.add('ustadAddFolderToLibrary',(folderName) => {
```
mikedawson, line 65: I would suggest you remove cy.contains("Library") from the command. This makes it impossible to use the command to create a subfolder.

The cy.contains("Library") click can be done in the test itself.

test-end-to-end/webapp-cypress/cypress/e2e/001_003_move_content.cy.js
```diff
@@ -0,0 +1,51 @@
...
+  .its('0.contentDocument')
+  .its('body')
+  .find(".h5p-question-check-answer.h5p-joubelui-button").click()
+  cy.go('back')
+ // Add new folder in library
+  cy.ustadAddFolderToLibrary('Test folder')
+ // Move content to new folder
+  cy.contains("Library").click()
+  cy.contains('Content_001').rightclick()
+  cy.contains("Move to",{timeout:3000}).click()
```
mikedawson: why do we need this timeout? Isn't the default timeout 4 seconds? Why make it one second lower? We should avoid using extra arguments when they are not required and the defaults work fine.

mikedawson (approved):

PR #851, opened 2024-02-26
Primary test maestro library 001

poojaustad:
Adding  content to Library - android  app


test-end-to-end/android-maestro/common/add_content.yaml
mikedawson, line 1: This is really clear_login_and_add_content (not only add_content) -  flow name should reflect it.

test-end-to-end/android-maestro/common/login_download_and_open_content.yaml
mikedawson, line 1: This flow attempt to play video content if that is available. I think that shouldn't be part of this flow. If the flow is login download and open, then that's what this flow should do. trying to select the play button should be part of the flow for video content.

test-end-to-end/android-maestro/e2e-tests/001_004_admin_can_move_content.yaml
```diff
@@ -0,0 +1,35 @@
...
+- tapOn: "New folder"
+- tapOn: "Title*"
+- inputText: "Test Folder"
+- tapOn: "SAVE"
+- takeScreenshot: "${TESTRESULTSDIR}/screenshots/before_moving_content_to_new_folder"
+- longPressOn:
+    text: "E-book"
+- tapOn: "Move"
+- tapOn: "Folder"
+- tapOn: "Move entries to this folder"
```
mikedawson, line 32: Before tapOn "Move entries to this folder" you need an assertion to check that you are now in the expected folder. 

I think you can use assertVisible on the title id - check that the title has changed to folder. Otherwise we risk the same thing as we had with cypress.

test-end-to-end/android-maestro/e2e-tests/001_004_admin_can_move_content.yaml
```diff
@@ -0,0 +1,35 @@
...
+- takeScreenshot: "${TESTRESULTSDIR}/screenshots/after_moving_content_to_new_folder_test"
+- tapOn: "Folder"
+- assertVisible: "E-book"
```
mikedawson, line 38: Here; again, we should have an assertion that the title has changed to make sure that we are in the expected folder. (e.g. after line 34 tapOn folder.

mikedawson (changes requested):

test-end-to-end/android-maestro/e2e-tests/001_004_admin_can_move_content.yaml
mikedawson: This was added:

assertVisible: "Test Folder"

But this won't have the intended effect. "Test Folder" needs to be in the title. The assert visible needs to include both text and the id of the title (app_title). 

This can pass if the folder hasn't opened yet, because "Test Folder" will still be in the list.

test-end-to-end/android-maestro/common/login_download_and_open_content.yaml
mikedawson: I still see this in login_download_and_open_content.yaml: 

```
- runFlow:
      when:
          visible:
            text: "Play"
      file: play_video_content.yaml
```

mikedawson (approved):

PR #869, opened 2024-04-03
Primary test maestro course blocks test checklist3

poojaustad:
Checklist 2 & 3


test-end-to-end/android-maestro/common/accepting_course_join_request.yaml
```diff
@@ -8,10 +8,8 @@ appId: com.toughra.ustadmobile
 -  tapOn: "Courses"
 - tapOn: ${TESTCOURSENAME}
 - tapOn: "MEMBERS"
+- scroll
```
mikedawson: This needs to be conditional - should use scrolluntil visible. Larger screen might not need scrolling (scroll might fail there)

test-end-to-end/android-maestro/common/add_module_block.yaml
```diff
@@ -6,18 +6,18 @@ appId: com.toughra.ustadmobile
...
-- tapOn: "Don't show before (Optional)"
-- tapOn: "OK"
+- tapOn: "Description"
+- tapOn:
+    id: "com.toughra.ustadmobile:id/editor"
+- inputText: "test block"
+- tapOn: "Done"
 - takeScreenshot: "${TESTRESULTSDIR}/screenshots/added_module_${MODULETITLE}"
-- tapOn: "SAVE"
+- tapOn: "Done"
```
mikedawson, line 26: This looks like tapping "done" twice in a row. We should add an assertion to make sure that the screen has changed (e.g. check the title has changed)

test-end-to-end/android-maestro/common/add_new_content_block_to_course.yaml
```diff
@@ -10,28 +10,23 @@ appId: com.toughra.ustadmobile
...
+- scrollUntilVisible:
+    element:
+      text: "Add block"
+    direction: DOWN
 # add new content block inside course
 - tapOn: "Add block"
 - tapOn: "Content"
-- tapOn: "Add new content"
-- tapOn: "Add file"
+- tapOn: "My content"
```
mikedawson: Do we need to tap "my content here"? Should not be required.

test-end-to-end/android-maestro/e2e-tests/002_004_course_title_is_mandatory.yaml
```diff
@@ -0,0 +1,30 @@
...
+- runFlow: "../common/clear_state_and_get_started.yaml"
+- runFlow: "../common/select_site.yaml"
+- runFlow:
+    file: "../common/user_login.yaml"
+- runFlow:
+    file: "../common/add_course.yaml"
+    env:
+      TESTCOURSENAME: " "
+# assert to make sure the course is not saved because of blank course title
+- assertVisible: "SAVE"
```
mikedawson: Yes the floating action button is not visible when editing, but I think it would be better to use the title for an assertion. Also we should make an assertion that the "required field" message appears.

test-end-to-end/android-maestro/e2e-tests/002_009_people_add_and_remove_profile_pic.yaml
```diff
@@ -0,0 +1,24 @@
...
+      TESTSERIAL: ${TESTSERIAL}
+      TESTNAME: ${TESTNAME}
+
+- runFlow:
+    file: "../common/add_profile_pic.yaml"
+    env:
+      TESTFILENAME: "admin_pic.jpg"
+      CONTROLSERVER: ${CONTROLSERVER}
+      TESTSERIAL: ${TESTSERIAL}
+- tapOn:
```
mikedawson, line 22: This needs an assertion that the picture is visible after being set e.g. open the profile and then make an assertion

mikedawson (approved):

PR #877, opened 2024-04-30
Added ustadTypeAndVerify

poojaustad:


test-end-to-end/webapp-cypress/cypress/e2e/WEB_004_001_assignment_creation_submission_grading.cy.js
```diff
@@ -64,7 +64,8 @@ it('Student submit assignment', () => {
   cy.contains('Assignment 1').click()
   cy.get('#assignment_text').get('div[contenteditable="true"]',{timeout:6000}).should('be.visible')
   cy.get('#assignment_text').click()
-  cy.get('#assignment_text').type("Text 1")
+  cy.get('#assignment_text').type("Text 1",{delay:25})
```
mikedawson: Why do we have type in addition to type and verify? We should use only type and verify here.

mikedawson (approved):

PR #895, opened 2024-06-24
Dev individual learner

Prashant-Android:
- Dispature.IO has been removed from all modules as it is not used in UstadAccountManager.kt.
- Individual screen components have been replaced with ListItem.
- Material Design 3 has been used.
- In App.kt, if(useBottomBar && currentSession?.endpoint?.isLocal != true) has been used.
- All string values are in lowercase.


core/src/commonMain/kotlin/com/ustadmobile/core/account/UstadAccountManager.kt
```diff
@@ -505,6 +526,56 @@ class UstadAccountManager(
...
+            db.userSessionDao.insertSession(newSession)
+
+            UserSessionWithPersonAndEndpoint(newSession, newPerson, fakeEndpoint)
+        }.also { newSession ->
+            addActiveEndpoint(fakeEndpoint)
+            currentUserSession = newSession
+        }
+    }
+
+    private fun randomString(length: Int): String {
```
mikedawson: We already have a randomString function - see RandomString.kt . Please use the existing function.

core/src/commonMain/kotlin/com/ustadmobile/core/impl/appstate/AppUiState.kt
```diff
@@ -24,7 +24,7 @@ data class AppUiState(
     val loadingState: LoadingUiState = LoadingUiState(),
     val title: String? = null,
     val navigationVisible: Boolean = true,
-    val hideBottomNavigation: Boolean = false,
+    var hideBottomNavigation: Boolean = false,
```
mikedawson: This should not be changed. AppUiState should be immutable in order to avoid side effects.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/OnBoardingViewModel.kt
```diff
@@ -53,14 +54,24 @@ class OnBoardingViewModel(
...
         settings[PREF_TAG] =  true.toString()
         navController.navigate(RedirectViewModel.DEST_NAME, buildMap {
             putFromSavedStateIfPresent(ARG_NEXT)
             putFromSavedStateIfPresent(ARG_OPEN_LINK)
         })
     }
 
+
+    fun onClickIndividual() {
+        navController.navigate(IndividualLearnerViewModel.DEST_NAME, buildMap {
```
mikedawson: Instead of using buildmap when this is empty, use emptyMap() (if this is required).

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/OnBoardingViewModel.kt
```diff
@@ -53,14 +54,24 @@ class OnBoardingViewModel(
...
+        })
+    }
+    fun onClickAddNewOrganization() {
+        //Redirect to website (ask mike that is there prebuild util class to open website)
```
mikedawson: This can be done using OpenExternalLinkUseCase . You can use the IDE to find usages of it. You can get it using the dependency injection.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/individual/IndividualLearnerViewModel.kt
```diff
@@ -0,0 +1,77 @@
...
+            accountManager.createLocalAccount()
+            goToNextDestAfterLoginOrGuestSelected(
+                accountManager.createLocalAccount().userSession.usPersonUid,
+                accountManager.createLocalAccount().endpoint
+            )
+        }
+    }
+
+    fun onClickRestoreFile() {
+        CoroutineScope(Dispatchers.Main).launch {
```
mikedawson: What is the coroutinescope doing here? 

This needs to take an argument for the file Uri.

The view will need to launch the file browser using the platform itself as is done in other places (see rememberUstadFilePickLauncher. For now, please just change the arguments and remove the (unused) CoroutineScope here.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/individual/IndividualLearnerViewModel.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        }
+    }
+
+    companion object {
+        const val DEST_NAME = "individualLearner"
```
mikedawson: DEST_NAME value should be capitalized e.g. IndividualLearner

core/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -160,6 +160,18 @@
     <string name="onboarding_headline3">Stay organized</string>
     <string name="onboarding_subheadline3">Education content, assignments, classes, and attendance all in one place</string>
     <string name="download_continue_stacked_label">Continue download</string>
+
+    <string name="onboarding_individual_text">I\'m an individual learner</string>
```
mikedawson: Please make the string match at least the first few words e.g. name="im_an_individual_learner". 

These are basically new strings. We don't use the prefixes anymore (because that makes it harder to find strings to reuse).

core/src/jsMain/kotlin/com/ustadmobile/core/MR.js.kt
```diff
@@ -0,0 +1,2245 @@
+package com.ustadmobile.core
```
mikedawson, line 1: This looks like something created by the IDE in response to a prompt. Please remove it.

This file will be generated for Javascript by the Moko gradle plugin when the Javascript build runs.

This file will break the javascript complation.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/components/UstadHorizontalDivider.kt
```diff
@@ -0,0 +1,28 @@
+package com.ustadmobile.libuicompose.components
```
mikedawson: This should be removed. Jetpack compose has a [HorizontalDivider](https://developer.android.com/develop/ui/compose/components/divider). We don't need to add our own

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/individual/IndividualLearnerScreen.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+
+}
+
+@Composable
+fun UstadTitleDescriptionButton(
+    title: String,
+    description: String,
+    titleColor: Color = Color.Black,
```
mikedawson: Avoid setting colors manually as far as possible. We should let the theme do its job.

With this code, if the user is in dark mode, this will go wrong.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/individual/IndividualLearnerScreen.kt
```diff
@@ -0,0 +1,112 @@
...
+
+
+        }
+    }
+
+
+}
+
+@Composable
+fun UstadTitleDescriptionButton(
```
mikedawson: I don't this function should exist. We can use ListItem as it is.

We should use the default color and sizes provided by the theme unless we have a real need to change them. That makes sure that everything works in dark mode, light mode, looks consistent, etc.

The clickable modifier should be applied to the list item for the click / ripple effect to work as expected.

lib-ui-compose/src/desktopMain/kotlin/com/ustadmobile/libuicompose/images/UstadImagePainter.kt
```diff
@@ -5,6 +5,7 @@ import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.res.painterResource
 
 val nameMap = mapOf(
+    UstadImage.GOOGLE_ICON to "/img/icn_google.png",
```
mikedawson: As we're not using the Google icon at this stage, let's remove that

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/individual/IndividualLearnerViewModel.kt
```diff
@@ -0,0 +1,83 @@
...
+        )
+    }
+
+    fun onClickContinueWithoutLogin() {
+        loadingState = LoadingUiState.INDETERMINATE
+
+        viewModelScope.launch {
+            accountManager.createLocalAccount()
+            goToNextDestAfterLoginOrGuestSelected(
+                accountManager.createLocalAccount().userSession.usPersonUid,
```
mikedawson: You're calling createLocalAccount three times. We want to create one account, not three.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/app/App.kt
```diff
@@ -118,6 +127,16 @@ fun App(
         }
     }
 
+    val isLocalSession = currentSession?.endpoint?.isLocal == true
```
mikedawson: Navigation should be done by the ViewModel (with very few exceptions e.g. if a login expires). This code could have side unforeseen side effects.

The individual user view model should be updated to simply go directly to ContentEntryListViewModel.DEST_NAME_HOME  and this section should be removed.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/onboarding/OnboardingScreen.kt
mikedawson, line 1: This UI should be updated to match the updated design as provided by Akshat e.g. the carousel should be removed. Choices should be in ListItem with the supporting text as per the design.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/individual/IndividualLearnerViewModel.kt
Prashant-Android: so now its look like this , 
  fun onClickContinueWithoutLogin() {
        loadingState = LoadingUiState.INDETERMINATE

        viewModelScope.launch {
            val localAccount = accountManager.createLocalAccount()
            goToNextDestAfterLoginOrGuestSelected(localAccount.userSession.usPersonUid, localAccount.endpoint)
        }
    }

core/src/androidMain/kotlin/com/ustadmobile/core/domain/backup/AndroidUnzipFileUseCase.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.ustadmobile.core.util.ZipProgress
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.withContext
+import java.io.File
+import java.io.FileInputStream
+import java.util.zip.ZipInputStream
+
+class AndroidUnzipFileUseCase : UnzipFileUseCase {
```
mikedawson: This should be in commonJvmMain : the zip classes on Android and JVM are the same.

core/src/commonMain/kotlin/com/ustadmobile/core/model/FileToZip.kt
```diff
@@ -0,0 +1,7 @@
+package com.ustadmobile.core.model
+
+
+data class FileToZip(
```
mikedawson: This data class should be within the UseCase interface - as per coding style doc.

core/src/commonMain/kotlin/com/ustadmobile/core/impl/di/CommonDomainModule.kt
```diff
@@ -30,6 +35,10 @@ import org.kodein.di.singleton
  * at the Application level di.
  */
 fun commonDomainDiModule(endpointScope: EndpointScope) = DI.Module("CommonDomain") {
+
+    bind<ZipFileUseCase>() with singleton { createZipFileUseCase() }
```
mikedawson: The DI binding for a UseCase which has different implementations on different platforms (or no implementation on certain platforms) should be done in platform specific DI - that is also now documented on the coding style guidelines. 

Please use the platform specific DI and remove the expect/actual for creation e.g. createZipFileUseCase etc.

core/src/androidMain/kotlin/com/ustadmobile/core/domain/backup/AndroidUnzipFileUseCase.kt
Prashant-Android: How will get this imports 

- import android.content.Context
- import android.net.Uri

mikedawson:
Closing as individual / personal mode functionality is now available, and included in dev-onboarding-simplification

PR #896, opened 2024-06-27
Invite UI

Nikunjsharma0:
invite via contact ui created from android / desktop


core/src/commonMain/kotlin/com/ustadmobile/core/domain/invite/ParseInviteUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+
+
+class ParseInviteUseCase {
+    var phoneNumValidatorUseCase: PhoneNumValidatorUseCase? = null
+    operator fun invoke(
+        text: String,
+        phoneValidatorUseCase: PhoneNumValidatorUseCase
+    ): List<InviteViaContactChip> {
+        phoneNumValidatorUseCase = phoneValidatorUseCase
+        val parts = text.split(",\\s*".toRegex())
```
mikedawson: Better to use .split(",").map { it.trim() } . Regex not needed here - and it can be tricky to predict exactly what a regex will do in any given edge case.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/invite/ParseInviteUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+                validatedChips.add(InviteViaContactChip(part, false))
+
+            }
+        }
+        return validatedChips
+
+    }
+
+    private fun isValidEmail(email: String): Boolean {
+        ValidateEmailUseCase().invoke(email)
```
mikedawson: ValidateEmailUseCase should also be a constructor parameter.

core/src/jvmTest/kotlin/com/ustadmobile/core/domain/invite/ParseInviteUseCaseTest.kt
mikedawson, line 1: This is a good start. Please add scenarios for invalid input.

lib-ui-compose/build.gradle.kts
```diff
@@ -122,6 +122,9 @@ android {
         jvmToolchain(17)
     }
 }
+//dependencies {
```
mikedawson: Please remove this. Generally we should not have commented out code in commits. If we need to refer to what the code was at some point before, then that is what Git is for. The only exception is when there is specific code that we would restore once some known issue is resolved.
Whenever we have commented out code, there should be a comment before it explaining why it is commented out instead of just removed.

lib-ui-compose/src/androidMain/kotlin/com/ustadmobile/libuicompose/components/UstadContactPickButton.kt
```diff
@@ -0,0 +1,158 @@
...
+    val pickContactLauncher = rememberLauncherForActivityResult(
+        contract = ActivityResultContracts.StartActivityForResult()
+    ) { result ->
+        if (result.resultCode == Activity.RESULT_OK) {
+            val data: Intent? = result.data
+            val contactUri: Uri? = data?.data
+            if (contactUri != null) {
+                val projection = arrayOf(ContactsContract.CommonDataKinds.Phone.NUMBER)
+                val cursor = context.contentResolver.query(contactUri, projection, null, null, null)
+                CoroutineScope(Dispatchers.IO).launch {
```
mikedawson: Instead of creating a new coroutine scope - use the coroutine scope that is provided from compose so that any task would be cancelled automatically if the user navigates away.

eg.

val scope = rememberCoroutineScope()

then to run something off the main thread
```
scope.launch {
   withContext(Dispatchers.IO) {
       //do something - will be cancelled automatically if user went away
   } 
}
```
scope.launch(Dispatchers.IO) might also work

lib-ui-compose/src/androidMain/kotlin/com/ustadmobile/libuicompose/components/UstadContactPickButton.kt
```diff
@@ -0,0 +1,158 @@
...
+                                it.getColumnIndex(ContactsContract.CommonDataKinds.Phone.NUMBER)
+                            val number = it.getString(numberIndex)
+
+                            onContactPicked(number)
+                        }
+                    }
+                }
+
+            } else {
+                onContactPicked(null)
```
mikedawson: If the user didn't select anything; then we don't need onContactPicked(null). 

If they did... but there was some kind of error (e.g. it.moveToFirst() doesn't have anything) then we should have a different function e.g. onContactError(t: Throwable?, message: String?). That can then go to the viewmodel, which can show an error using showSnackBar

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/clazz/inviteViaContact/InviteViaContactScreen.kt
```diff
@@ -0,0 +1,147 @@
...
+class AvatarChip(text: String, val avatar: ImageVector) : Chip(text)
+
+@Composable
+fun InviteViaContactScreen(
+    uiState: InviteViaContactUiState = InviteViaContactUiState(),
+    onChipSubmitClick: (String) -> Unit,
+) {
+    val state = rememberChipTextFieldState<AvatarChip>()
+    val scrollState = rememberScrollState()
+    Column(
```
mikedawson: Please use UstadVerticalScrollColumn - it takes care of showing the scrollbar itself on Desktop.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/clazz/inviteViaContact/InviteViaContactScreen.kt
```diff
@@ -0,0 +1,147 @@
...
+            onSubmit = {
+
+                onChipSubmitClick(it)
+
+                null
+
+            },
+
+            chipStyle = ChipTextFieldDefaults.chipStyle(shape = RoundedCornerShape(20.dp)),
+            textStyle = TextStyle(fontSize = 13.sp),
```
mikedawson: Please use a MaterialTheme for text style.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/clazz/inviteViaContact/InviteViaContactScreen.kt
```diff
@@ -0,0 +1,147 @@
...
+            onContactPicked = {
+                if (it != null) {
+                    onChipSubmitClick(it)
+                }
+            }
+        )
+    }
+
+    LaunchedEffect(uiState.chips) {
+        uiState.chips?.let { validChip ->
```
mikedawson: Bad variable name: you don't know its a valid chip at this line.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/clazz/inviteViaContact/InviteViaContactScreen.kt
```diff
@@ -0,0 +1,147 @@
...
+        UstadContactPickButton(
+            onContactPicked = {
+                if (it != null) {
+                    onChipSubmitClick(it)
+                }
+            }
+        )
+    }
+
+    LaunchedEffect(uiState.chips) {
```
mikedawson, line 125: I don't think this function will work as intended. Every time the uiState chips change it will add it again

e.g. user adds chip A, goes to view model, comes back on ui State, gets added to chip state [uistate = [A], chip state = [A].
user adds chip B, goes to view model, come back on ui state as [A,B], both get added, now chip state [ A, A, B]

mikedawson: This needs to check the material chips state to see if the given chip is already there before adding it.

mikedawson: Also I don't see code handling the removal.

mikedawson (changes requested):
Generally

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/clazz/inviteViaContact/InviteViaContactScreen.kt
Nikunjsharma0: we dont neet to worry about removal of chips because when we remove any chip this state = rememberChipTextFieldState<AvatarChip>()  will remove chip on it own

mikedawson: Yes we do need to worry about removal: when the user actually goes and clicks invite, then the viewmodel needs to know (otherwise the invite request will include items removed by the user).

Another reason that state needs to be in the ViewModel: we use the saved state handle (in the view model) e.g. if the app is destroyed when the user leaves due to inactivity, and then returns, we should restore the state.

mikedawson:
This pull request is now obsolete since it has been merged into dev-simplified-onboarding, and will be merged to primary through that.

PR #908, opened 2024-07-16
Offline installation and sharing functionality.

Prashant-Android:
- Implemented offline sharing capability.
- Updated relevant components to handle offline data sharing.


core/src/androidMain/kotlin/com/ustadmobile/core/domain/share/SendAppFileUseCaseAndroid.kt
```diff
@@ -0,0 +1,40 @@
...
+                throw IllegalArgumentException("APK has splits and cannot be shared as a single file.")
+            }
+
+            val intent = Intent(Intent.ACTION_SEND).apply {
+                type = "application/vnd.android.package-archive"
+                putExtra(Intent.EXTRA_STREAM, apkUri)
+                addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+            }
+
+            val chooserIntent = Intent.createChooser(intent, "Share App")
```
mikedawson: "Share App" - this needs to be added to the string xml. 

Catch: AppContext MUST be the activity, not the application context (otherwise localization will be wrong). Bind the use case in the activity, not in UstadApp.

mikedawson:
When the user clicks "Share the app" show a bottom sheet with choices: "Share app store link", "Send app file (APK)"

If user selects Share app store link, just use the package name to make the link

mikedawson:
Add a buildconfig field (e.g. buildconfig.default.properties) com.ustadmobile.googleplayurl set to the Ustad Mobile Google Play URL.

Then update app-android/build.gradle to add this to manifest placeholders

Then add this as a metadata proeprty to AndroidManifest (same as is done for other proeprties e.g. com.ustadmobile.presetlocale

app-android/src/main/res/xml/file_provider_path.xml
```diff
@@ -15,4 +15,7 @@
     <files-path
         name="files"
         path="." />
-</paths>
\ No newline at end of file
+    <root-path
```
mikedawson, line 18: Why has this been added?

docs/source/installation.rst
```diff
@@ -0,0 +1,86 @@
...
+
+1. On the device with the app, open the app and go to "Share App" in settings.
+2. Select "QuickShare" as the sharing method.
+3. On the receiving device, open QuickShare and scan for nearby devices.
+4. Select the sending device from the list of available devices.
+5. Accept the file transfer on both devices.
+6. Once the transfer is complete, tap on the received APK to install.
+
+.. image:: images/sharing1.jpg
+   :alt: Alternate text describing the image
```
mikedawson: Please set the alt text here e.g. Screenshot of (insert description of what is being shown on screenshot picture)

core/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -4,6 +4,10 @@
     <!--suppress CheckTagEmptyBody -->
     <string name="blank" translatable="false"></string>
     <string name="login">Login</string>
+    <string name="share_app_title">Share App</string>
```
mikedawson: Strings names should be as per the text itself e.g. share_app not share_app_title (see coding style docs). Please make the string names match the text used for all strings added. 

Please also screenshot where the text is used and save the screenshot in translate/screenshots (file name should match the string name used in xml). Use an image editor (e.g. Gimp) to highlight the text within the screenshot (as per the other screenshots).

This is used in our translation platform so that translators can see a screenshot of the word in context so they can translate it properly.

core/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -4,6 +4,10 @@
...
+    <string name="share_app_description">Share App with Others</string>
```
mikedawson: Text should be "Share this app with others", which would make the name "share_this_app_with_others".

mikedawson:
@Prashant-Android please put the share option in the account screen (below 'add another account, above the version'). I can't see this option in any other app, so its hard to say exactly where to put it.

I think it makes sense to put this option in the same place where anything 'about app' is located - which is mostly in the account / profiles sections of all the other apps that I can find.

img/individual_create.svg
mikedawson, line 1: These images are not part of offline sharing - they should not be in this pull request.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/settings/SettingsViewModel.kt
```diff
@@ -117,7 +122,13 @@ class SettingsViewModel(
 
     private val settings: Settings by instance()
 
+    private val sendAppFileUseCase: SendAppFileUseCase by instance()
```
mikedawson: This would **crash** all non-Android platforms. This must be changed to use instanceOrNull. If it is null, then the share option should not be visible.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/share/SendAppFileUseCase.kt
```diff
@@ -0,0 +1,5 @@
+package com.ustadmobile.core.domain.share
+
+interface SendAppFileUseCase {
```
mikedawson: Let's rename this to ShareAppUseCase because it is now having both the send file option and share app link.

docs/source/installation.rst
```diff
@@ -0,0 +1,86 @@
...
+   :width: 400px
+   :height: 700px
+
+.. image:: images/sharing3.jpg
+   :alt: Alternate text describing the image
+   :width: 400px
+   :height: 700px
+
+Other Sharing Apps
```
mikedawson, line 87: Please add a link for offline sharing using Google Play. We don't need to copy from Google, just link eg. Title 

**Share offline using Google Play**

You can share the app offline using Google Play as per [Google Play instructions](https://support.google.com/googleplay/answer/9283534?hl=en).

app-android/src/main/res/xml/file_provider_path.xml
Prashant-Android: The <root-path> element tells the FileProvider where to find the root directory of our app’s private storage. Without this, the FileProvider wouldn’t know where to look, and we wouldn’t be able to share the APK file.

mikedawson: If that was the case, why did everything else work without it (e.g. opening submission assignment files)? I don't think it is needed. Please test removing it.

core/src/commonMain/resources/MR/base/strings.xml
mikedawson: Should be "Share **a**pp" NOT "Share App"

app-android/src/main/res/xml/file_provider_path.xml
Prashant-Android: I tested without this it's not working.

Failed to find configured root that contains / data/app/~~08Rm8hFVcAc05pWaKjKlyg==/ com.toughra.ustadmobile-08fWnSG-ZblrNg0W HQoPzQ==/base.apk

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/accountlist/AccountListViewModel.kt
```diff
@@ -91,7 +93,11 @@ class AccountListViewModel(
     private val dontSetCurrentSession: Boolean = savedStateHandle[ARG_DONT_SET_CURRENT_SESSION]
         ?.toBoolean() ?: false
 
+    private val shareAppUseCase: ShareAppUseCase? by instanceOrNull()
+    val sendAppOptionVisible = shareAppUseCase != null
```
mikedawson: This is **not** following MVVM pattern. 

MVVM works as per coding style / docs: the UiState object has properties that are then observed by the view. shareAppOptionVisible should be a property of the AccountListUiState - set by the viewmodel - like all other properties.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/accountlist/AccountListViewModel.kt
```diff
@@ -143,6 +149,18 @@ class AccountListViewModel(
         }
     }
 
+    fun onClickAppShare(shareLink: Boolean) {
+        viewModelScope.launch {
+            try {
+                shareAppUseCase?.invoke(shareLink)
+            } catch (e: IllegalArgumentException) {
```
mikedawson: The catch clauses are exactly the same: you only need one catch e. Best practice now is to use Throwable instead of Exception.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/settings/SettingsViewModel.kt
```diff
@@ -39,11 +40,14 @@ data class SettingsOfflineStorageOption(
 )
 
 data class SettingsUiState(
-
+    val sendAppOptionVisible: Boolean = false,
```
mikedawson: Properties related to backup etc. are not related to this pull request, should not be here. sendAppOptionVisible is no longer used here (now it is moved to accountlist), so should also be removed from this uiState.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/settings/SettingsScreen.kt
```diff
@@ -97,43 +104,72 @@ fun SettingsScreen(
...
             }
         }
     }
 
-    if(uiState.waitForRestartDialogVisible) {
+    if (uiState.waitForRestartDialogVisible) {
         UstadWaitForRestartDialog()
     }
 
+    if (shareAppOptionsVisible) {
```
mikedawson: Needs removed from here and should be only in accountlistscreen.

core/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -4,6 +4,10 @@
     <!--suppress CheckTagEmptyBody -->
     <string name="blank" translatable="false"></string>
     <string name="login">Login</string>
+    <string name="share_app">Share app</string>
+    <string name="share_this_app_with_others">Share This App with Others</string>
```
mikedawson: This string (share_this_app_with_others) is no longer used, like anything else that is unused, it should be removed.

docs/source/installation.rst
mikedawson: This is not resolved.

docs/source/installation.rst
```diff
@@ -0,0 +1,98 @@
...
+
+1. Copy the APK file to your SD card or flash drive.
+2. Insert the SD card or connect the flash drive to your Android device.
+3. Use a file manager app to navigate to the APK file.
+4. Tap on the APK file to begin the installation process.
+5. If prompted, allow installation from unknown sources in your device settings.
+6. Follow the on-screen instructions to complete the installation.
+
+.. image:: images/installation1.jpg
+   :alt: Alternate text describing the image
```
mikedawson: These items still need alt text.

mikedawson (approved):

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/accountlist/AccountListScreen.kt
```diff
@@ -138,24 +168,34 @@ fun AccountListScreen(
             )
         }
 
+        item(key = "share_app") {
```
mikedawson: The share app option should only appear if uiState.showShareAppOption is true. This needs to go inside an if statement.

Right now - it will show on Desktop, but won't work there.

mikedawson (changes requested):
Show share app option needs to be within if condition to avoid it showing on unsupported platforms (e.g. desktop)

mikedawson (approved):

PR #918, opened 2024-07-23
Add image file

poojaustad:


test-end-to-end/android-maestro/common/add_profile_pic.yaml
```diff
@@ -15,6 +15,11 @@ appId: com.toughra.ustadmobile
 # clear the downloads
 - evalScript: ${http.get(CONTROLSERVER + "cleardownloads?device=" + TESTSERIAL)}
 
+# clear the downloads
```
mikedawson: Why do we need to call cleardownloads twice in a row?

mikedawson (approved):

PR #986, opened 2024-11-26
Implement ValidateUsernameUseCase for Username Validation



Prashant-Android:
This PR introduces the ValidateUsernameUseCase to enforce consistent username validation across the app, addressing the following:

Allowable characters: letters (including non-English), numbers, . and _.
Disallowed characters: whitespace and banned special symbols (e.g., !@#$%^&*()).
Error message updated for clarity.


core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/person/edit/PersonEditViewModel.kt
```diff
@@ -356,29 +359,7 @@ class PersonEditViewModel(
     }
 
     private fun validateUsername(username: String): Boolean {
```
mikedawson: This function can be deleted and the validate username use case can be called directly. This function serves no purpose.

mikedawson:
This validation needs to be used by PersonAccountEditViewModel - which is where the admin would go to create an account.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/validateusername/ValidateUsernameUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+     * - Non-English Unicode characters are allowed
+     * - Must be 3–15 characters long after trimming
+     *
+     * @param username The username string to validate
+     * @return The valid username (trimmed) if valid, or null if invalid
+     */
+    operator fun invoke(username: String): String? {
+        val trimmed = username.trim()
+
+        if (trimmed.length !in 3..15) {
```
mikedawson: Clear violation of coding style rule: "never hardcode literals" see https://github.com/UstadMobile/UstadMobile/blob/primary/CODING-STYLE.md#never-hardcode-any-literal-values

core/src/commonMain/kotlin/com/ustadmobile/core/domain/validateusername/ValidateUsernameUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+class ValidateUsernameUseCase {
+
+    /**
+     * Validates a username according to the following rules:
+     * - Must not contain spaces, tabs, or banned special characters like !@#$%^&*()
+     * - Non-English Unicode characters are allowed
+     * - Must be 3–15 characters long after trimming
+     *
+     * @param username The username string to validate
+     * @return The valid username (trimmed) if valid, or null if invalid
```
mikedawson: If banned characters are found, then this should remove/replace them with and return whatever is valid 

add an argument: ```invalidReplacement: String = ""```

Any banned characters should then be replaced with the replacement. This is going to be required when running the migration to fix any invalid usernames found.

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/ext/UmAppDatabaseMigrations.kt
```diff
@@ -1653,6 +1653,43 @@ val MIGRATION_201_202 = DoorMigrationStatementList(201, 202) { db ->
     }
 }
 
+val MIGRATION_202_203_SERVER = DoorMigrationStatementList(202, 203) { db ->
```
mikedawson, line 1656: The migration needs to use the ValidateUsernameUseCase so it follows the don't repeat yourself principle -e.g. the logic for what is, and is not, a valid username, should be in one place.

If you put this migration into ```jvmMain``` source set, you can then use ```db.connection``` to get the underlying JDBC database connection so you can run any JDBC query or statement.

It should go through all usernames in batches, check the username, and then if a username is found where the return result of validateusernameusecase is not the same as the username (e.g. something in the username is invalid), it should be updated (e.g. the update statement should only run on the usernames that need changed).

The migration as-below has different logic. It would not find white space other than space (e.g. won't find tab, carriage return, etc). It also won't find or fix usernames with special characters.

mikedawson: Note: you may need to put the server side migration in app-ktor-server or core/jvmMain source set, otherwise, it won't have access to the ValidateUsernameUseCase.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/validateusername/ValidateUsernameUseCase.kt
```diff
@@ -0,0 +1,69 @@
...
+        val containsInvalidChars = lowercased.any { !isCharacterAllowed(it) }
+        if (containsInvalidChars && invalidReplacement.isEmpty()) {
+            return null
+        }
+
+        // Replace invalid characters if replacement provided
+        return if (containsInvalidChars) {
+            lowercased.map { char ->
+                when {
+                    !isCharacterAllowed(char) -> invalidReplacement
```
mikedawson: What's going on here is two separate logic tracks - one dealing with white space, and another other non-allowed characters? Why not just have isCharacterAllowed return false when character is white space? Why have different logic handling white space vs. other non-allowed characters?

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/login/LoginViewModel.kt
```diff
@@ -156,9 +159,26 @@ class LoginViewModel(
...
 
-    fun onUsernameChanged(username: String) {
-        _uiState.update { prev ->
-            prev.copy(username = username)
+    fun onUsernameChanged(newValue: String) {
+        // If no input or input is valid (including whitespace)
+        val lastChar = newValue.lastOrNull()
+        val isFirstChar = newValue.length == 1
+
+        val isValid = when {
```
mikedawson: This is repeating the same logic that is in the use case. That's against [coding style](https://github.com/UstadMobile/UstadMobile/blob/primary/CODING-STYLE.md#dont-repeat-yourself) - "Don't repeat yourself" -important principle in software development ( see also https://en.wikipedia.org/wiki/Don%27t_repeat_yourself ).

There is now separate logic validating the username in three places now: PersonAccountEdit, Login, and the ValidateUsernameUseCase. That is very problematic : for all reasons mentioned in the Wikipedia article e.g. more likely to have a bug, if something needs changed it needs changed in multiple places. Same concept as I referenced in Meshrabiya for converting ip addresses to Int etc.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/login/LoginViewModel.kt
```diff
@@ -156,9 +159,26 @@ class LoginViewModel(
         }
     }
 
-    fun onUsernameChanged(username: String) {
-        _uiState.update { prev ->
-            prev.copy(username = username)
+    fun onUsernameChanged(newValue: String) {
+        // If no input or input is valid (including whitespace)
+        val lastChar = newValue.lastOrNull()
```
mikedawson: This logic fails when anytime the user has changed any character which is not at the end of the text field.

mikedawson:
Closing as this is now replaced with #999

PR #998, opened 2024-12-17
Dev simplified onboarding clean

Nikunjsharma0:
Added passkey changes and invite user branch merged


app-react/src/jsMain/kotlin/com/ustadmobile/view/clazz/inviteViaContact/ClazzInviteRedeem.kt
```diff
@@ -22,7 +22,7 @@ import web.cssom.px
 
 
 external interface ClazzInviteRedeemProps : Props {
-    var uiState: InviteRedeemUiState
+    var uiState: ClazzInviteUiState
```
mikedawson: Should use the full base name for UiState class - ClazzInviteRedeemUiState

app-android/src/main/java/com/ustadmobile/port/android/impl/UstadApp.kt
```diff
@@ -1079,6 +1080,14 @@ class UstadApp : Application(), DIAware, ImageLoaderFactory{
             )
         }
 
+        bind<ClazzRedeemUseCase>() with scoped(LearningSpaceScope.Default).provider {
```
mikedawson: Use case should also be ClazzInviteRedeemUseCase - naming should be consistent.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/redirect/RedirectViewModel.kt
```diff
@@ -87,7 +87,7 @@ class RedirectViewModel(
                     put(
                         ARG_LEARNINGSPACE_URL,
                         apiUrlConfig.presetLearningSpaceUrl?.requireHttpPrefix()
-                            ?.requirePostfix("/")?.replace("/umapp", "") ?: ""
+                            ?.requirePostfix("/")?: ""
```
mikedawson: The /umapp shouldn't have been there - it's not part of the learning space url ever. 

For style:

if you used:
```
val presetLearningSpaceUrl = apiUrlConfig.presetLearningSpaceUrl
```

Then you could remove null safety within the if statement. That should be done

mikedawson (changes requested):

mikedawson (approved):

PR #999, opened 2024-12-23
Filter out invalid characters from a username

### LINKED ISSUE #982: Apply validation for username
[](https://devserver3.ustadmobile.com/jenkins/job/dev-filter-username/)

[](https://devserver3.ustadmobile.com/jenkins/job/dev-filter-username-End-To-End-Tests/)

### **User Story:**
As an **User**, I want the **Username field to be validated** so that **any human errors can be avoided**

- [x] @Prashant-Android  Create new branch and pull request.
- [x] @poojaustad Create Jenkins job to build automatically on commit and nightly end-to-end build of branch
- [x] @poojaustad Update end-to-end testing scenarios in accordance with user story
- [x] @poojaustad Add Cypress and Maestro tests to branch based on testing scenarios/wireframes
- [x] @Prashant-Android  Implement code as per design and acceptance criteria: requires successful Jenkins build and end-to-end build. Post video of completion of acceptance criteria on task card. Self check for at least 5-10 minutes.
- [ ] @poojaustad check/verify - run any manual sanity checks required

Implementation: this should be a domain use case e.g. ValidateUsernameUseCase

### **Background info:**

- Currently the username field allows space, special characters etc
- We need to add validation to the field so that we can avoid users entering special characters, spaces , numbers etc

### **Acceptance criteria:**
- All screens where username is entered - login screen, create account screen should:
- Block special characters (except ```.``` and ```_```  and whitespace (space, tab, return etc)). When the user is using a keyboard - the key event itself should be blocked.
- Block starting a username with a number. A number can come at the end. Starting with a number prevents other screens  (eg invite screen) from differentiating between a username and a phone number 
- Convert any upper case letters to lower case.

Prashant-Android:
Filter out invalid characters from a username


core/src/commonMain/kotlin/com/ustadmobile/core/domain/filterusername/FilterUsernameUseCase.kt
```diff
@@ -0,0 +1,35 @@
...
+                char.isDigit() && username.firstOrNull() != char -> char
+                else -> invalidCharReplacement
+            }
+        }.joinToString("")
+    }
+
+    companion object {
+        private val VALID_USERNAME_SPECIAL_CHARS = setOf('.', '_')
+
+        fun isCharAllowed(char: Char): Boolean = when {
```
mikedawson: This is named as basically the **opposite of what it actually does** : it returns true for a character that is not allowed, and false for a character that is allowed. Please ensure meaningful names as per [pull request checklist](https://github.com/UstadMobile/UstadMobile/blob/primary/PULLREQUEST-CHECKLIST.md#coding-style) and basic good practice

mikedawson (changes requested):

sivedcodes:
Android

[Android.webm](https://github.com/user-attachments/assets/08b7e870-d24b-4669-9493-43add38864a3)

sivedcodes:
Desktop

[Desktop.webm](https://github.com/user-attachments/assets/55a23a71-22e9-4ce9-a5c6-e3d25aaa1f8f)

app-react/src/jsMain/kotlin/UstadJsDi.kt
```diff
@@ -81,6 +82,10 @@ internal fun ustadJsDi(
         )
     }
 
+    bind<FilterUsernameUseCase>() with singleton {
```
mikedawson, line 86: ValidateUsernameUseCase must also be bound in DI for web.

Both use cases need bound in app-desktop

app-react/src/jsMain/kotlin/com/ustadmobile/view/login/LoginScreen.kt
```diff
@@ -74,6 +75,14 @@ private val LoginComponent2 = FC<LoginProps> { props ->
                 onTextChange = {
                     props.onUsernameValueChange(it)
                 }
+                onKeyDown = { event ->
+                    val char = event.key.singleOrNull()
+                    if (char != null) {
```
mikedawson: Should be shorter: does not need two if statements.

```
if(char != null && FilterUsernameUseCase.shouldBlockKeyEvent(char)) {
   event.preventDefault()
}
```

app-react/src/jsMain/kotlin/com/ustadmobile/view/person/accountedit/PersonAccountEditScreen.kt
```diff
@@ -42,6 +42,12 @@ val PersonAccountEditComponent2 = FC<PersonAccountEditScreenProps> { props ->
                     helperText = ReactNode(props.uiState.usernameError ?: strings[MR.strings.required])
                     error = props.uiState.usernameError != null
                     disabled = !props.uiState.fieldsEnabled
+                    onKeyDown = { event ->
+                        val char = event.key.singleOrNull()
+                        if (char != null) {
```
mikedawson: As above - should be shorter

app-react/src/jsMain/kotlin/com/ustadmobile/view/person/edit/PersonEditScreen.kt
```diff
@@ -242,6 +242,12 @@ val PersonEditComponent2 = FC <PersonEditScreenProps> { props ->
                     disabled = !props.uiState.fieldsEnabled
                     error = props.uiState.usernameError != null
                     helperText = ReactNode(props.uiState.usernameError ?: strings[MR.strings.required])
+                    onKeyDown = { event ->
```
mikedawson, line 246: As above - should be shorter / use one if statement

core/src/commonMain/kotlin/com/ustadmobile/core/domain/validateusername/ValidateUsernameUseCase.kt
```diff
@@ -0,0 +1,41 @@
...
+        return when {
+            username.length < MIN_LENGTH -> ValidationResult.INVALID_TOO_SHORT
+            username.length > MAX_LENGTH -> ValidationResult.INVALID_TOO_LONG
+            username.firstOrNull()?.isDigit() == true -> ValidationResult.INVALID_STARTS_WITH_NUMBER
+            !username.all { it.isValidUsernameChar() } -> ValidationResult.INVALID_OTHER
+            else -> ValidationResult.VALID
+        }
+    }
+
+    private fun Char.isValidUsernameChar(): Boolean = when {
```
mikedawson: shouldBlockKeyEvent =  NOT (isValidUsernameChar)

So why are there two functions with the same logic, **_AGAIN_** (same problem as per [comment on previous pull request](https://github.com/UstadMobile/UstadMobile/pull/986))?

Have one function in ValidateUsernameUseCase as follows:

```
companion object {
    fun isValidUsernameChar(character: Char) {
         return when {
             character.isLetter() -> true
             character.isDigit() -> true
             character in ALLOWED_SPECIAL -> true
             else -> false 
         }
    }

}
```

Remove ```FilterUsernameUseCase.shouldBlockKeyEvent``` - replace this with using ```!ValidateUsernameUseCase.isValidUsernameChar ```

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/person/accountedit/PersonAccountEditScreen.kt
```diff
@@ -52,19 +57,22 @@ fun PersonAccountEditScreen(
...
                 },
-                onValueChange = {
-                    onChange(
-                        uiState.personAccount?.copy(
-                            username = it
-                        )
-                    )
-
+                onValueChange = { newValue ->
+                    newValue.lastOrNull()?.let { lastChar ->
```
mikedawson: I already pointed out on [previous pull request](https://github.com/UstadMobile/UstadMobile/pull/986) - _"This logic fails when anytime the user has changed any character which is not at the end of the text field."_ - comment on LoginViewModel.kt

Please remove all value change event processing based on the flawed assumption that the last character in the string is the one that was just changed.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/person/edit/PersonEditScreen.kt
```diff
@@ -221,16 +226,25 @@ fun PersonEditScreen(
...
                 enabled = uiState.fieldsEnabled,
                 isError = uiState.usernameError != null,
                 singleLine = true,
-                onValueChange = {
-                    onPersonChanged(uiState.person?.shallowCopy{
-                        username = it
-                    })
+                onValueChange = { newValue ->
+                    val lastChar = newValue.lastOrNull()
+                    if (lastChar == null || !FilterUsernameUseCase.shouldBlockKeyEvent(lastChar)) {
```
mikedawson: I already pointed out on https://github.com/UstadMobile/UstadMobile/pull/986 - "This logic fails when anytime the user has changed any character which is not at the end of the text field." - comment on LoginViewModel.kt

Please remove all value change event processing based on the flawed assumption that the last character in the string is the one that was just changed.

mikedawson (changes requested):

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/person/accountedit/PersonAccountEditScreen.kt
mikedawson: Use the filter username use case on change. If the username is valid - filter username use case won't change it. If something invalid has been entered via IME (or copy/paste) - then this will catch it.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/filterusername/FilterUsernameUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Filters and normalizes username input:
+ * - Converts uppercase to lowercase
+ * - Filters invalid characters
+ * - Handles key event validation
+ */
+class FilterUsernameUseCase {
+    operator fun invoke(
+        username: String,
+        invalidCharReplacement: Char
```
mikedawson: invalidCharReplacement should probably be changed to a string; because when it is being used by the viewmodel, the replacement for an invalid character is blank (empty string).

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/login/LoginViewModel.kt
```diff
@@ -156,10 +162,13 @@ class LoginViewModel(
         }
     }
 
-    fun onUsernameChanged(username: String) {
-        _uiState.update { prev ->
-            prev.copy(username = username)
-        }
+    fun onUsernameChanged(newValue: String) {
+        val filteredValue = filterUsernameUseCase(
```
mikedawson, line 167: See above - we could avoid the extra .filter {} call by changing the invalidCharReplacement type to string, so then you can use ```invalidCharReplacement = ""```

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/person/edit/PersonEditViewModel.kt
```diff
@@ -288,9 +294,20 @@ class PersonEditViewModel(
     }
 
     fun onEntityChanged(entity: Person?) {
+        val filteredEntity = entity?.let { currentEntity ->
```
mikedawson: The logic you have here is correct; that we should only filter the username if it was changed. I think this would be clearer if it starts with the if statement, and I think that can go into the existing update e.g.

```
_uiState.update { prev -> 
    prev.copy(
        person = if(entity?.username != _uiState.value.person?.username) {
            entity?.shallowCopy { 
                username = filterUsernameUseCase(
                     username = entity.username ?: "",
                     invalidCharReplacement = ""
               )
            } 
        }else {
             entity
        }  
    )
}
```

core/src/commonMain/kotlin/com/ustadmobile/core/domain/validateusername/ValidateUsernameUseCase.kt
```diff
@@ -0,0 +1,41 @@
...
+ * Validates whether a username meets all required criteria:
+ * - Must not be too short or too long
+ * - Must not start with a number
+ * - Must only contain valid characters (letters, numbers, dots, underscores)
+ */
+class ValidateUsernameUseCase {
+
+    enum class ValidationResult {
+        VALID,
+        INVALID_TOO_SHORT,
```
mikedawson: I think we need to show an error message specific to the validation result: e.g. if the username is rejected because it is too short, we should show only "Invalid: must be at least 3 characters". Many of our users are a bit unfamiliar with tech, so I think it would be better to tell them what is wrong with the username as they entered it, rather than make them read all the rules and figure it out. 

You can add a StringResource to the ValidationResult (e.g. ```ValidationResult(val errorMessage: StringResource?)```).

Then in viewModel, you can use it e.g.

```
val validationResult = validateUsernameUseCase(...)
if(validationrResult != Valid) {
    usernameError = validationResult.errorMessage?.let { systemImpl.get(it} }
}

mikedawson (changes requested):
Please also ensure that both use cases are bound on all three platforms: including Desktop (this was noted on last review).

mikedawson:
@Prashant-Android the Jenkins [build fails](https://devserver3.ustadmobile.com/jenkins/job/dev-filter-username/1/) due to failing tests: please fix this ASAP.

Sived-Codes:
Screenshot_20250102-120042
Screenshot_20250102-120203
Screenshot_20250102-120324

mikedawson (approved):

PR #1000, opened 2024-12-24
saving password in password manager

Nikunjsharma0:
saving password in password manager


core/src/commonMain/kotlin/com/ustadmobile/core/domain/password/SavePasswordUseCase.kt
```diff
@@ -0,0 +1,5 @@
+package com.ustadmobile.core.domain.password
+
+interface SavePasswordUseCase {
```
mikedawson, line 6: I think this should have comments to explain that this is saving the password to a password manager (e.g. not being saved internally). The interface name itself is fine.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/password/LoginWithSavedPasswordUseCase.kt
```diff
@@ -0,0 +1,5 @@
+package com.ustadmobile.core.domain.password
+
+interface LoginWithSavedPasswordUseCase {
```
mikedawson: We can't make two separate requests to credential manager - one for saved password, then another for a passkey.

We need to have one ```GetCredentialUseCase``` - which could return with either a passkey or a saved username/password. The return result can be a sealed class (which would be either a passkey or a username/password).

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/signup/SignupEnterUsernamePasswordViewModel.kt
```diff
@@ -272,6 +274,10 @@ class SignupEnterUsernamePasswordViewModel(
 
                 }
 
+                savePasswordUseCase?.invoke(
+                    username =savePerson.username.toString(),
+                    password = _uiState.value.password.toString())
```
mikedawson, line 279: Indentation - final closing bracket should be on a new line as per Kotlin stlye e.g.
```
savePasswordUseCase?.invoke(
    username =savePerson.username.toString(),
    password = _uiState.value.password.toString()
)
```

lib-ui-compose/src/androidMain/kotlin/com/ustadmobile/libuicompose/util/password/LoginWithSavedPasswordUseCaseImpl.kt
```diff
@@ -0,0 +1,38 @@
...
+
+            val credentialResponse = credentialManager.getCredential(
+                context = context,
+                request = GetCredentialRequest(
+                    credentialOptions = listOf(GetPasswordOption())
+                )
+            )
+
+            val credential = credentialResponse.credential as? PasswordCredential
+            return LoginWithSavePasswordResult(
```
mikedawson: FYI: Closing bracket needs to be on new line as per Kotlin style, although this use case itself will be deleted.

lib-ui-compose/src/androidMain/kotlin/com/ustadmobile/libuicompose/util/password/SavePasswordUseCaseImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+                context = context,
+                request = CreatePasswordRequest(
+                    id = username,
+                    password = password
+                )
+            )
+            Napier.d { "Password saved successfully for user: $username" }
+        } catch (e: CreateCredentialNoCreateOptionException) {
+            Napier.w { "No option to create credentials: ${e.message}" }
+        } catch (e: CreateCredentialException) {
```
mikedawson: When logging exceptions, pass the exception to the logger so it can log the stack trace unless it is an common exception that has no major effect.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/login/LoginViewModel.kt
```diff
@@ -167,7 +163,16 @@ class LoginViewModel(
                 }
             }
         }
+
         onSignInWithPassKey()
+        viewModelScope.launch {
+            val result=  loginWithSavedPasswordUseCase?.invoke()
+            if (result?.username != null&& result.password != null){
```
mikedawson: Couldn't this just be if ```result != null```?

mikedawson (changes requested):

mikedawson:
Also: UseCase implementation should not be in libuicompose - should be in the androidMain source directory of core.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/password/LoginWithSavedPasswordUseCase.kt
mikedawson: LoginWithSavedPasswordUseCase is no longer used and should be removed. We should also remove ```LoginWithSavePasswordResult``` which is not used anywhere else.

lib-ui-compose/src/androidMain/kotlin/com/ustadmobile/libuicompose/util/passkey/GetCredentialUseCaseImpl.kt
mikedawson, line 1: As per [previous comment on this pull request](https://github.com/UstadMobile/UstadMobile/pull/1000#issuecomment-2563393791) all use cases (and implementations) should be in the core module _not_ the lib-ui-compose module.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/passkey/GetCredentialUseCase.kt
```diff
@@ -0,0 +1,6 @@
+package com.ustadmobile.core.domain.passkey
+
+interface GetCredentialUseCase {
```
mikedawson, line 3: It would be better for this usecase to follow the same pattern as other use cases which are related to a given learning space (e.g. enrollment, invite, etc) - let's use the DI scope with LearningSpace as a parameter, not use domain as a parameter to the invoke function.

e.g.


```interface GetCredentialUseCase

suspend operator fun invoke(): CredentialResult
```

Then in the implementation can use something like this:
```
class GetCredentialUseCaseImpl(
   private val learningSpace: LearningSpace
) {
    private val domain: String by lazy {
       Url(learningSpace.url).host
    }
     
   
    operator fun invoke() {
       //logic here    
    }  
}

```

lib-ui-compose/src/androidMain/kotlin/com/ustadmobile/libuicompose/util/passkey/GetCredentialUseCaseImpl.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.json.JSONObject
+
+class GetCredentialUseCaseImpl(
+    val context: Context,
+    val passkeyRequestJsonUseCase: PasskeyRequestJsonUseCase
+) : GetCredentialUseCase {
+    override suspend fun invoke(domain: String): CredentialResult {
+        val credentialManager = CredentialManager.create(context)
+
+        val getPasswordOption = GetPasswordOption()
```
mikedawson: Get Password option needs parameters to ensure that we only show passwords for the given learning space url.

mikedawson (changes requested):

mikedawson:
When using getCredentials - once it has been bound to DI using the learning space scope - you'll need to manually use the learning space as specified by server url:

e.g. 
```
val credentialUseCase: GetCredentialUseCase  = di.on(LearningSpace(serverUrl)).instanceOrNull()
```

mikedawson (approved):

mikedawson (approved):

PR #1006, opened 2024-12-30
Dev reports option editing

### LINKED ISSUE #992: Customizable reports: option editing
[](https://devserver3.ustadmobile.com/jenkins/job/dev-reports-option-editing/)

[](https://devserver3.ustadmobile.com/jenkins/job/dev-reports-option-editing-End-To-End-Tests/)

**User story**

As a _teacher/decision maker/parent_ I want _customize reports_ so that _I can track progress_.

User Background info on user story:

Technical notes:
* The report options are stored as a data class that can be serialized to a string (the string itself will be saved on a database entity)
* A domain UseCase (separate task) must then convert the report options into valid SQL that will generate the requested report

**Tasks**

- [x] @mikedawson review model class structure
- [x] @mikedawson [wireframe](https://balsamiq.cloud/sagy57t/podvg7r/rF904) draft completion
- [x] @poojaustad create Jenkins branch for regular build and end-to-end build
- [x] @poojaustad write (human readable) step-by-step test description
- [x] @poojaustad add Cypress and Maestro test for description as above (expected 11/Feb)
- [x] @Anugraha-sutara update model classes to use enums and sealed classes instead of Int constants
- [x] @Anugraha-sutara implement ReportEdit as per wireframe
- [x] @Anugraha-sutara implement ReportFilterEdit as per wireframe
- [x] @Anugraha-sutara implement conditions to ensure user cannot select incompatible y axis types
- [x] @Anugraha-sutara branch to build successfully pass all existing tests on Jenkins and tests for acceptance criteria
- [x] @poojaustad  update end-to-end test such that it should only login and create the report, not actually use the content.

**Acceptance criteria**

* User should be able to create a new report with options as per the data classes ( see [Report2.kt](https://github.com/UstadMobile/UstadMobile/blob/dev-reports-mvvm/core/src/commonMain/kotlin/com/ustadmobile/core/domain/report/model/Report2.kt) )

Behavior notes:
* Subgrouping needs to have an option for no subgrouping (default)

Anugraha-sutara:
- update model classes to use enums and sealed classes instead of Int constants
- implement ReportEdit as per wireframe
- implement ReportFilterEdit as per wireframe


core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
mikedawson, line 1: Package naming should follow the same conventions as others: this should be in com.ustadmobile.core.viewmodel.report.__edit__

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+    companion object {
+        const val DEST_NAME = "Report"
+        const val DEST_NAME_HOME = "ReportHome"
+        val ALL_DEST_NAMES = listOf(DEST_NAME, DEST_NAME_HOME)
+        const val RESULT_KEY_REPORT = "arg"
+        const val ENTITY_UID = "uid"
+    }
+}
+
+data class ReportEditUiState(
```
mikedawson: Please put ReportEditUiState at the top like as is done with all other view models.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+        )
+    }
+
+
+    fun onAddSeries() {
+        _uiState.update { prev ->
+            prev.copy(
+                reportOptions2 = prev.reportOptions2.copy(
+                    series = prev.reportOptions2.series + ReportSeries2(
+                        reportSeriesUid = prev.reportOptions2.series.size + 1,
```
mikedawson: This way of setting reportSeriesUid is flawed: if a user added series, then removed them, there is no gaurantee that you would not have duplicates.

You could use:

```
reportOptions2.series.maxOf { it.reportSeriesUid } + 1
```

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+                hideBottomNavigation = false
+            )
+        }
+        launchIfHasPermission(
+            setLoadingState = true,
+            permissionCheck = { true }
+        ) {
+            async {
+                loadEntity(
+                    serializer = Report.serializer(),
```
mikedawson: The Serializer here should be ```ReportOptions2.serializer()``` because that is the type we are using on the uistate.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+class ReportEditViewModel(
+    di: DI,
+    savedStateHandle: UstadSavedStateHandle,
+    destName: String = ""
+) : UstadEditViewModel(di, savedStateHandle, DEST_NAME) {
+
+    private val _uiState: MutableStateFlow<ReportEditUiState> =
+        MutableStateFlow(ReportEditUiState())
+    val uiState: Flow<ReportEditUiState> = _uiState.asStateFlow()
+    private val entityUid: Long
```
mikedawson: You don't need your own ```entityUid``` val. You can get this from the parent class's ```entityUidArg```. I think this was in PersonEditViewModel because it was created before entityUidArg was added to the parent class.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import org.kodein.di.DI
+
+class ReportEditViewModel(
+    di: DI,
+    savedStateHandle: UstadSavedStateHandle,
+    destName: String = ""
```
mikedawson: destName should never be blank (could even cause navigation problems) - you can use the DEST_NAME as default.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+                        _uiState.update { prev ->
+                            prev.copy(
+                                reportOptions2 = loadedReport?.reportOptions?.takeIf { it.isNotBlank() }
+                                    ?.let { Json.decodeFromString(ReportOptions2.serializer(), it) }
+                                    ?: ReportOptions2()
+                            )
+                        }
+                    }
+                )
+                navResultReturner.filteredResultFlowForKey(RESULT_KEY_REPORT).collect { result ->
```
mikedawson: Result returner should not be used like this. The ReportFilterEditViewModel should use receive and return only the ReportFilter2, not anything else.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+                        println("Report options updated successfully: ${report}")
+                    }
+                } catch (e: Exception) {
+                    println("Error updating report options: ${e.message}")
+                }
+            }
+        }
+    }
+
+    fun onEntityChanged(newOptions: ReportOptions2) {
```
mikedawson: onEntityChanged __must__ use ```scheduleEntityCommitToSavedState``` so that changes made by the user are saved to the saved state handle - which is required to ensure data is not lost when the app is destroyed / recreated and  on the web.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/ReportFilterEditViewModel.kt
mikedawson, line 1: This needs to be moved to the correct package e.g. ```com.ustadmobile.core.viewmodel.report.filteredit```

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/ReportFilterEditViewModel.kt
```diff
@@ -1,37 +1,104 @@
...
+import com.ustadmobile.core.impl.nav.UstadSavedStateHandle
+import com.ustadmobile.core.viewmodel.report.ReportEditViewModel
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.update
+import org.kodein.di.DI
 
 data class ReportFilterEditUiState(
+    val filters: ReportFilter3? = ReportFilter3(),
+    val filterCondition: ReportFilter2? = null
```
mikedawson: The condition should not be a separate field: it is part of the filter.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/report/reportedit/ReportEditScreen.kt
```diff
@@ -0,0 +1,331 @@
...
+                Text(
+                    text = stringResource(MR.strings.add_series),
+                )
+            }
+        }
+    }
+}
+
+@Composable
+fun EditReportDropdown(
```
mikedawson: Please do **not** create a new dropdown component (EditReportDropDown must be removed). Even UstadExposedDropDownField itself shouldn't really exist.

Please implement dropdowns as per the [Jetpack Compose docs for ExposedDropdownMenuBox](https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#ExposedDropdownMenuBox(kotlin.Boolean,kotlin.Function1,androidx.compose.ui.Modifier,kotlin.Function1)) .

If you see there is duplicated logic (e.g. the  isExpanded state), then you can a component to avoid repeated logic.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/report/reportedit/ReportEditScreen.kt
```diff
@@ -0,0 +1,331 @@
...
+    label: String,
+    options: List<MessageIdOption2>,
+    onOptionSelected: (MessageIdOption2) -> Unit,
+    modifier: Modifier = Modifier,
+    isError: Boolean = false,
+    enabled: Boolean = true,
+    supportingText: (@Composable () -> Unit)? = null,
+) {
+    Column(modifier = Modifier) {
+        Text(label, fontWeight = FontWeight.SemiBold)
```
mikedawson: As I have mentioned previously: always use the theme and **never** set styles manually unless it is explicitly noted otherwise in the task card or prototype. We must **always** use the theme to ensure that dark mode, light mode, user font size settings, etc are handled properly and manually setting fonts / styles at best will make the application look inconsistent, at worse will start breaking things. 

This is as per the [coding style](https://github.com/UstadMobile/UstadMobile/blob/primary/CODING-STYLE.md#screens) - "Screens must use default margins and colors provided by the platform theme (Jetpack Compose or MUI) unless specifically noted otherwise"

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ReportExt.kt
mikedawson, line 1: Please delete this code instead of commenting it out. If we need to refer to this code, we can look at primary.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    fun onClickSave() {
+        viewModelScope.launch {
+            val currentReport = _uiState.value.reportOptions2
+            val report = Report(
+                reportTitle = currentReport.title,
+                reportOptions = Json.encodeToString(currentReport),
+            )
+            scheduleEntityCommitToSavedState(report, serializer = Report.serializer())
```
mikedawson: scheduleEntityCommitToSavedState should be in the onEntityChanged function, and run almost immediately when the user changes anything.

The edit view models work as follows:

1) As the user modifies an entity through the UI form etc - those changes are put into the saved state handle. This way - any changes they made are saved in case the app is destroyed and recreated, and changes can be saved on the web if the user navigates backwards and forwards.

Changes are not (yet) saved in the database. The user might not want to save those changes, they can still discard those changes if they made mistakes etc.

2) Once the user clicks save, then the entity is saved into the database.

The Serializer type here should be ReportOptions2

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
```diff
@@ -0,0 +1,230 @@
...
+            activeRepo.withDoorTransactionAsync {
+                try {
+                    if (entityUid == 0L) {
+                        activeRepo.reportDao().insert(report)
```
mikedawson: You need to use insertAsync and updateAsync. If you try and use a non-async database call inside an async transaction it will probably deadlock (get stuck).

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/report/filteredit/ReportFilterEditScreen.kt
mikedawson, line 1: Please make sure that the package names for all screens match exactly the patterns as per their related viewmodel.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/report/model/ReportFilter2.kt
```diff
@@ -1,57 +1,81 @@
...
-    var reportFilterValue: String? = null,
-
-    var reportFilterDropDownValue: Int = 0,
-
-    var reportFilterValueBetweenX: String? = null,
-
-    var reportFilterValueBetweenY: String? = null,
-) {
+/** Enum representing different comparison types */
+enum class Comparisons(val value: Int) {
```
mikedawson: The StringResource should be a field on the Comparison Enum e.g.
```
enum class Comparison(val value: Int, val stringResource: StringResource)
```

No need to create the map, extra function, etc.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/report/model/ReportFilter2.kt
```diff
@@ -1,57 +1,81 @@
...
+    var reportFilterField: FilterType? = null,
 
-        const val CONDITION_IN_LIST = 206
+    var reportFilterCondition: Comparisons? = null,
 
-        const val CONDITION_NOT_IN_LIST = 207
+    var reportFilterValue: String? = ""
+)
 
+fun getComparisonSymbol(comparison: Comparisons): String {
```
mikedawson: Again - there is no need to add this extra function. The comparison symbol can just be an extra field on the enum class.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/edit/ReportEditViewModel.kt
```diff
@@ -0,0 +1,242 @@
...
+                                reportOptions2 = loadedReport ?: ReportOptions2()
+                            )
+                        }
+                    }
+                )
+
+                launch {
+                    navResultReturner.filteredResultFlowForKey(RESULT_KEY_REPORT)
+                        .collect { result ->
+                            val reportResult = result.result as? Map<*, *> ?: return@collect
```
mikedawson, line 81: No need to use a map as the result type. ReportFilterEdit should set the result as the ReportFilter3 itself. Then collect it. No need for the map.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/filteredit/ReportFilterEditViewModel.kt
```diff
@@ -0,0 +1,106 @@
...
+            entityUid = filter?.reportFilterUid?.toLong() ?: 0,
+            result = mapOf(
+                "filter" to filter?.copy(reportFilterSeriesUid = seriesId),
+            )
+        )
+    }
+
+
+    fun onEntityChanged(value: ReportFilter3?) {
+        _uiState.update { currentState ->
```
mikedawson: This also needs to use schedulecommit to saved state.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/report/ReportEditViewModel.kt
mikedawson: This is still an issue. There is no need to use a map. ReportFilterEditViewModel should 'return' ReportFilter3 as the result, and that should be then be collected. No need for a map (same as other screens).

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/ReportFilterEditViewModel.kt
mikedawson: This is not yet resolved: why do we have ReportFilter3 and ReportFilter2 here?

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/report/edit/ReportEditScreen.kt
```diff
@@ -0,0 +1,355 @@
...
+                            label = null,
+                            singleLine = true,
+                            onValueChange = { newTitle ->
+                                val updatedSeries = seriesItem.copy(reportSeriesTitle = newTitle)
+                                onSeriesChanged(updatedSeries)
+                            },
+                        )
+                        Icon(
+                            imageVector = Icons.Filled.Close,
+                            contentDescription = "Remove filter",
```
mikedawson: contentDescriptions are used by text-to-speech tools - so they must use localized strings (eg stringResource) the same as text that appears 'normally'.

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/report/edit/ReportEditScreen.kt
```diff
@@ -0,0 +1,355 @@
...
+            options = options,
+            selectedValue = selectedValue,
+            onOptionSelected = onOptionSelected
+        )
+    }
+}
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun ExposedDropdownMenu(
+    options: List<MessageIdOption2>,
```
mikedawson: This shouldn't need to use MessageIdOption2. 

I recommend:
1) Create an interface e.g. OptionWithLabel
```
interface OptionWithLabelStringResource {
    val label: StringResource
}
```

2. Update the enum classes to implement the interface above

3. the ExposedDropDownMenu function will then be much easier/cleaner e.g.

```
fun <T: OptionWithLabelStringResource> ExposedDropdownMenu(
   options: List<T>,
   selectedValue: T,
   onOptionSelected: (T) -> Unit
)
```

Then there will be no need to map to/from integers etc.

core/src/commonMain/kotlin/com/ustadmobile/core/impl/locale/entityconstants/ReportXaxisConstants.kt
```diff
@@ -0,0 +1,15 @@
+package com.ustadmobile.core.impl.locale.entityconstants
+
+import com.ustadmobile.core.MR
+import com.ustadmobile.core.domain.report.model.ReportXAxis
+import com.ustadmobile.core.util.MessageIdOption2
+
+object ReportXAxisConstants {
```
mikedawson: This should be possible to remove.

core/src/commonMain/kotlin/com/ustadmobile/core/util/ext/ReportExt.kt
mikedawson: Pending

mikedawson (approved):

PR #1061, opened 2025-03-20
Dev copy course

Mandvii:
Added Copy Button if permission is present
App bar Title updating according to option clicked
Navigating to Edit Screen when Copy is clicked (app bar title changed -"Copy course")


app-react/src/jsMain/kotlin/com/ustadmobile/view/clazz/detailoverview/ClazzDetailOverviewScreen.kt
```diff
@@ -100,18 +105,31 @@ val ClazzDetailOverviewComponent = FC<ClazzDetailOverviewProps> { props ->
...
                             direction = responsive(StackDirection.row)
-
-                            if(props.uiState.managePermissionVisible) {
-                                UstadQuickActionButton {
-                                    text = strings[MR.strings.permissions]
-                                    icon = ShieldIcon.create()
-                                    onClick = {
-                                        props.onClickPermissions()
+                            if (props.uiState.managePermissionVisible) {
+                                Stack {
```
mikedawson, line 109: This is an extra Stack. It should not exist. The buttons are already in a stack. Please remove.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/clazz/list/ClazzListViewModel.kt
```diff
@@ -161,17 +165,23 @@ class ClazzListViewModel(
     }
 
     override fun onClickAdd() {
-        navigateToCreateNew(ClazzEditViewModel.DEST_NAME)
+        navigateToCreateNew(
+            ClazzEditViewModel.DEST_NAME,
+            extraArgs = mapOf(UstadView.ARG_CAN_ADD_COURSE to _uiState.value.canAddNewCourse.toString())
```
mikedawson, line 170: No! 

ClazzEditViewModel should check permissions to determine what to show - just like every other view model does. It should not be passed from anywhere else. **Revert this change**.

This makes the code less predictable. Adding an extra argument that needs passed to ClassEdit, from every which path someone could get to ClazzEdit. This is against the single point of truth principle.

It's also insecure: someone could simply add the parameter themselves to the URL.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/clazz/list/ClazzListViewModel.kt
```diff
@@ -161,17 +165,23 @@ class ClazzListViewModel(
...
+                UstadView.ARG_CODE_TABLE to Clazz.TABLE_ID.toString()))
     }
-
     fun onClickEntry(entry: Clazz) {
-        navigateOnItemClicked(ClazzDetailViewModel.DEST_NAME, entry.clazzUid, entry)
+        navigateOnItemClicked(
+            ClazzDetailViewModel.DEST_NAME,
+            entry.clazzUid,
+            entry,
+            mapOf(UstadView.ARG_CAN_ADD_COURSE to _uiState.value.canAddNewCourse.toString())
```
mikedawson, line 183: As mentioned above - use of UstadView.ARG_CAN_ADD_COURSE must be removed **entirely**.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/UstadEditViewModel.kt
```diff
@@ -69,11 +73,10 @@ abstract class UstadEditViewModel(
     ) {
         val popUpToViewName = savedStateHandle[UstadView.ARG_RESULT_DEST_VIEWNAME]
         val saveToKey = savedStateHandle[UstadView.ARG_RESULT_DEST_KEY]
-
         val createdNewEntity = savedStateHandle[ARG_ENTITY_UID] == null
         val returnResultExpected = (popUpToViewName != null && saveToKey != null)
-
-        if(createdNewEntity && !returnResultExpected) {
+        val isCopyAction = detailViewExtraArgs[UstadView.ARG_ACTION_TYPE] == ClazzEditViewModel.COPY
```
mikedawson, line 78: NO!

UstadEditViewModel (a common parent class) should NOT be changed with logic that is specific to one specific view model.

Revert all your changes to UstadEditViewModel. **This pull request must NOT make any changes to UstadEditViewModel**.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/clazz/detail/ClazzDetailViewModel.kt
```diff
@@ -39,7 +39,8 @@ class ClazzDetailViewModel(
         val tabs = mutableListOf(
             TabItem(
                 viewName = ClazzDetailOverviewViewModel.DEST_NAME,
-                args = mapOf(UstadView.ARG_ENTITY_UID to entityUidArg.toString()),
+                args = mapOf(UstadView.ARG_ENTITY_UID to entityUidArg.toString(),
+                    UstadView.ARG_CAN_ADD_COURSE to canAddCourse.toString()),
```
mikedawson, line 43: As mentioned: all use of UstadView.ARG_CAN_ADD_COURSE must be removed

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/clazz/detailoverview/ClazzDetailOverviewViewModel.kt
```diff
@@ -110,6 +112,7 @@ class ClazzDetailOverviewViewModel(
     )
 
     val listRefreshCommandFlow: Flow<RefreshCommand> = _listRefreshCommandFlow.asSharedFlow()
+    private val canAddNewCourse: Boolean = savedStateHandle[UstadView.ARG_CAN_ADD_COURSE]?.toBoolean() ?: false
```
mikedawson, line 115: As mentioned: this is the wrong approach. You need to collect the flow / make the permission check within ClazzDetailOverviewViewModel (as is done by all other viewmodels). You must not rely on an argument coming in from other view models for this.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/clazz/detailoverview/ClazzDetailOverviewViewModel.kt
```diff
@@ -244,7 +251,7 @@ class ClazzDetailOverviewViewModel(
             ARG_CLAZZUID to entityUidArg.toString(),
         )
 
-        when(courseBlock.cbType) {
+        when (courseBlock.cbType) {
```
mikedawson, line 254: Please do not change spacing or reformat when no other change is being made. It just makes the pull request harder to review.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/clazz/detailoverview/ClazzDetailOverviewViewModel.kt
```diff
@@ -275,9 +286,11 @@ class ClazzDetailOverviewViewModel(
         }
     }
 
-    private fun onClickEdit() {
-        navController.navigate(ClazzEditViewModel.DEST_NAME,
-            mapOf(UstadView.ARG_ENTITY_UID to entityUidArg.toString()))
+    fun onClickEdit(actionType: String) {
```
mikedawson, line 289: NO!

For two reasons:
1) This is clearly against the ViewModel event pattern : for any event, there should be a function onClickEdit means the user clicked edit. You should add a separate onClickCopy function.
2) Even if that was going to be done, a string parameter is the wrong way to handle that. A string can have an infinite number of values. In this case there are only two valid values (so if one was doing it that way, one should use a boolean or enum)

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/clazz/edit/ClazzEditViewModel.kt
```diff
@@ -392,17 +427,19 @@ class ClazzEditViewModel(
 
     fun onCheckedAttendanceChanged(checked: Boolean) {
         onEntityChanged(_uiState.value.entity?.shallowCopy {
-            clazzFeatures = if(checked) {
+            clazzFeatures = if (checked) {
```
mikedawson, line 430: As mentioned - please don't change the spacing when nothing else has changed - just makes code review harder.

mikedawson (commented):
Needs a completely new start. Too many issues with this pull request. Not reviewable.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/UstadEditViewModel.kt
Mandvii: Step 1 -i click on course items (in ClazzListScreen)
Step 2- navigate to course detail page and I click on Copy button  (in ClazzDetailOverviewScreen)
Step 3- navigate to Edit Screen then add details and click save (ClazzEditScreen)
Step 4- After saving it goes to ClazzDetailOverviewScreen  that shows the previous course details which was clicked instead of updating the  screen with the copied course details.
So I added a if else condition in the function being used.

So for that which function can be used or else any other function need to be written

mikedawson:
Closing as this needed redone as per review comments

PR #1084, opened 2025-05-12
username suggestions added

Nikunjsharma0:
Dev simplified onboarding username into signup


core/src/commonMain/kotlin/com/ustadmobile/core/domain/credentials/passkey/webAuthn/PasskeyWebAuthNResponse.kt
```diff
@@ -0,0 +1,15 @@
+package com.ustadmobile.core.domain.credentials.passkey.webAuthn
+
+import kotlinx.serialization.Serializable
+
```
mikedawson: This needs a comment with a link to the specification; this is a model based on an external specification.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/signup/OtherSignUpOptionSelectionViewModel.kt
```diff
@@ -115,13 +115,7 @@ class OtherSignUpOptionSelectionViewModel(
 
             val passkeyCreated = createPasskeyUseCase?.invoke(
                 CreatePasskeyParams(
```
mikedawson: Why does the create passkey params class still need to exist? The only parameter is username.

Also : do not use .toString on a nullable string. This is just hiding when something is null that shouldn't be the case. null.toString = "null" - that is not what we want.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/signup/SignUpViewModel.kt
```diff
@@ -248,8 +270,69 @@ class SignUpViewModel(
...
+    fun onUsernameChanged(newValue: String) {
+        val filteredValue = filterUsernameUseCase(
+            username = newValue,
+            invalidCharReplacement = ""
+        )
+
+        val updatedPerson = _uiState.value.person?.shallowCopy {
+            username = filteredValue
+        }
+        if (_uiState.value.person?.username!=filteredValue){
```
mikedawson: Here it would be better to do one _uiState.update call, and put the if statement in that. 

e.g.

_uiState.update { prev -> 
  prev.copy(
     usernameSetByUser = if(...),
     person = ...  
}

Everytime you call update - that will trigger compose to recompose. Better to only do that once.

gradle/libs.versions.toml
```diff
@@ -59,6 +59,7 @@ generateuuid =  "0.8.4"
 urlencoder =  "1.4.0"
 nestedscroll-webview =  "0.1.6"
 androidx-webkit =  "1.10.0"
+codegen =  "0.120"
```
mikedawson: should be json-kotlin-schema-codegen - not just "codegen"

core/src/commonMain/kotlin/com/ustadmobile/core/domain/credentials/passkey/webAuthn/PasskeyWebAuthNResponse.kt
```diff
@@ -0,0 +1,16 @@
...
+import kotlinx.serialization.Serializable
+//during registration and signin with passkey the standard webAuthN response
+// received as per https://w3c.github.io/webauthn/#dictdef-registrationresponsejson
+@Serializable
+data class PasskeyWebAuthNResponse(
+    val authenticatorAttachment: String? = null,
+    val clientExtensionResults: ClientExtensionResults? = null,
+    val id: String,
+    val rawId: String? = null,
```
mikedawson: As per spec - what is required should not be nullable.

core/src/commonMain/kotlin/com/ustadmobile/core/domain/credentials/passkey/webAuthn/PasskeyWebAuthNResponse.kt
```diff
@@ -0,0 +1,16 @@
+package com.ustadmobile.core.domain.credentials.passkey.webAuthn
+
+import kotlinx.serialization.Serializable
+//during registration and signin with passkey the standard webAuthN response
+// received as per https://w3c.github.io/webauthn/#dictdef-registrationresponsejson
+@Serializable
+data class PasskeyWebAuthNResponse(
```
mikedawson: Class should be named as per the spec -  RegistrationResponseJSON

mikedawson: .. same applies to other classes e.g. response class is AuthenticatorAttestationResponseJSON etc.

mikedawson (approved):

PR #1107, opened 2025-05-29
admin can handle bottom nav visibility

Nikunjsharma0:
to control the visibility of bottom nav items


lib-database/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/Site.kt
```diff
@@ -60,9 +64,12 @@ class Site {
 
     var authSalt: String? = null
 
+    var bottomNavVisibilityFlag: Long = 0
+
     companion object {
 
         const val TABLE_ID = 189
+        const val BOTTOM_NAV_DEFAULT_FLAG = 15L
```
mikedawson, line 72: Don't use constant 15L - this needs to use the flags with OR

mikedawson:
Replaced with PR # 1116

PR #1116, opened 2025-06-10
admin can handle bottom nav visibility

### LINKED ISSUE #1105: Toggle Feature for Bottom Navigation in Admin Settings
**Jenkins Status**

 [](https://devserver3.ustadmobile.com/jenkins/job/primary-bottomnav-visiblitysetting/)

[](https://devserver3.ustadmobile.com/jenkins/job/primary-bottomnav-visiblitysetting-End-To-End-Tests/)

- [ ] @mikedawson to check apparently unrelated assignment regression

**User story**

As an admin, I want to have toggle options in the admin settings for showing or hiding the bottom navigation items (Course, Library, Messages, and People), so that I can customise the user navigation experience based on the organisation’s needs.

Design is updated [here](https://balsamiq.cloud/sagy57t/podvg7r/r36F6)

Nikunjsharma0:
to control the visibility of bottom nav items


lib-database/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/Site.kt
```diff
@@ -60,10 +64,15 @@ class Site {
 
     var authSalt: String? = null
 
+    var bottomNavVisibilityFlag: Long? = null
```
mikedawson: Add ColumnInfo annotation (including default value) - you should use the real default value.

lib-database/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/Site.kt
```diff
@@ -60,10 +65,16 @@ class Site {
 
     var authSalt: String? = null
 
+    @ColumnInfo(defaultValue = "0")
+    var bottomNavVisibilityFlag: Long = 0
```
mikedawson, line 65: Constants for flags should be on this companion object

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/BottomNavVisibilityFlags.kt
```diff
@@ -0,0 +1,10 @@
+package com.ustadmobile.core.db
+
+object BottomNavVisibilityFlags {
```
mikedawson: This class should be removed - flags should be in Site.kt

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/app/App.kt
```diff
@@ -54,12 +58,30 @@ import moe.tlaster.precompose.navigation.rememberNavigator
...
 data class TopNavigationItem(
     val destRoute: String,
     val icon: ImageVector,
     val label: StringResource,
 )
+fun getVisibleTopLevelNavItems(
+    bottomNavVisibilityFlag: Long?
+): List<TopNavigationItem> {
+    return APP_TOP_LEVEL_NAV_ITEMS.filter {
+        when (it.destRoute) {
```
mikedawson: Flag should be added to the data class TopNavigationItem

lib-database/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/Site.kt
mikedawson: Should be an int - 32 potential values is more than enough

lib-ui-compose/src/commonMain/kotlin/com/ustadmobile/libuicompose/view/app/App.kt
mikedawson: Please apply the same approach on the web version

app-react/src/jsMain/kotlin/com/ustadmobile/mui/components/Sidebar.kt
```diff
@@ -32,14 +34,30 @@ import web.cssom.px
...
+
 }
 
 data class RootScreen(
     val key: String,
     val nameMessageId: StringResource,
     val icon: FC<*>? = null,
 )
-
+fun getVisibleRootScreens(bottomNavVisibilityFlag: Long?): List<RootScreen> {
```
mikedawson: Should use same approach as per comment on App.kt (compose)

app-react/src/jsMain/kotlin/com/ustadmobile/mui/components/UstadScreens.kt
```diff
@@ -100,6 +102,13 @@ val UstadScreens = FC<Props> {
     val langConfig = useMemo(dependencies = emptyArray()) {
         loaderData.di.direct.instance<SupportedLanguagesConfig>()
     }
+    val currentDb: UmAppDatabase =  loaderData.di.direct.on(accountManager.activeLearningSpace)
+        .instance(tag = DoorTag.TAG_DB)
+
+    val currentSite by currentDb.siteDao().getSiteAsFlow().collectAsState(null)
```
mikedawson: Pass the current site as a parameter (like app.kt) not the screen list

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/site/edit/BottomNavPermissionConstants.kt
```diff
@@ -0,0 +1,14 @@
+package com.ustadmobile.core.viewmodel.site.edit
+import com.ustadmobile.core.MR
+import com.ustadmobile.core.db.BottomNavVisibilityFlags
+
+object BottomNavPermissionConstants {
+
+    val BOTTOM_NAV_LABELS = listOf(
```
mikedawson: Should be removed as per comments on flags

app-react/src/jsMain/kotlin/com/ustadmobile/view/site/edit/SiteEditScreen.kt
```diff
@@ -25,18 +29,20 @@ import react.FC
...
+    var onTogglePermission: (Long) -> Unit
     var onChangeTermsLanguage: (UstadMobileSystemCommon.UiLanguage) -> Unit
     var onChangeTermsHtml: (String) -> Unit
 }
 
 val SiteEditComponent2 = FC<SiteEditProps> { props ->
 
     val strings = useStringProvider()
-
+    val theme by useRequiredContext(ThemeContext)
```
mikedawson: The IDE is saying its unused - so it shouldn't be here. Don't ignore warnings on commit

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/site/edit/SiteEditViewModel.kt
```diff
@@ -158,7 +167,20 @@ class SiteEditViewModel(
             }
         }
     }
-
+    fun onTogglePermission(flag: Long) {
+        _uiState.update { prev ->
+            val entityVal = prev.site
```
mikedawson: No need for this if/else - can and should be prev.site?.copy()

lib-database/src/commonMain/kotlin/com/ustadmobile/core/db/ext/UmAppDatabaseMigrations.kt
```diff
@@ -1692,15 +1693,23 @@ val MIGRATION_206_207 = DoorMigrationStatementList(206, 207) { db ->
     }
 }
 
+
 val MIGRATION_207_208 = DoorMigrationStatementList(207, 208) { db ->
+
```
mikedawson, line 1707: This migration number needs to be bumped, and previous 207-208 migration should be here.

lib-database/src/commonMain/kotlin/com/ustadmobile/lib/db/entities/Site.kt
mikedawson: Kept as long because other functions to handle permission constants etc use Long type.

app-react/src/jsMain/kotlin/com/ustadmobile/view/site/edit/SiteEditScreen.kt
mikedawson: Still - has unused imports (can be removed by simply using ctrl+alt+O)

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/site/edit/BottomNavPermissionConstants.kt
mikedawson: Add a new file: SiteConstants (in com.ustadmobile.viewmodel.site ). 

That should have a list of the top level items info that is common (e.g. the route name, the label string constant, the flag constant).

Then on each platform - create a list of those common constants plus the icon 

```
val iconMap = mapOf(
 flag to icon
) 
val TOP_LEVEL = commonTopLevel.map { ...
  it to iconMap[it.flag]
}
```

mikedawson:
To be modified: GetDefaultDestinationUseCase to choose where to go based on what is enabled. This can be changed to using a suspend function, so it run a database query and get the site entity from database.

app-desktop/src/main/java/com/ustadmobile/port/desktop/App.kt
```diff
@@ -205,6 +211,11 @@ fun main() {
             val showPoweredBy = remember {
                 di.direct.instance<GetShowPoweredByUseCase>().invoke()
             }
+            val accountManager: UstadAccountManager = di.direct.instance()
+            val currentDb: UmAppDatabase = di.direct.on(accountManager.activeLearningSpace)
+                .instance(tag = DoorTag.TAG_DB)
+
+            val currentSite by currentDb.siteDao().getSiteAsFlow().collectAsState(initial = null)
```
mikedawson: This doesn't handle when the learning space changes. You need to use the flow from AccountManager to get the current learning space. 

e.g. if you use something like:

var currentSite: Site? by mutableStateOf(null)

LaunchedEffect(Unit) {
   accountManager.currentSession.map {
     it.learningSpace
  }.distinct(). collectLatest {
      val db = di.on(session.learningSpace).instance<Database>
      db.siteDao.collectSiteAsFlow().collect {
           currentSite = it
      }
   }
}

app-react/src/jsMain/kotlin/com/ustadmobile/mui/components/Sidebar.kt
```diff
@@ -1,52 +1,60 @@
...
-import com.ustadmobile.core.viewmodel.contententry.list.ContentEntryListViewModel
-import com.ustadmobile.core.viewmodel.message.conversationlist.ConversationListViewModel
-import com.ustadmobile.core.viewmodel.person.list.PersonListViewModel
+import com.ustadmobile.core.viewmodel.site.COMMON_TOP_LEVEL_NAV_ITEMS
+import com.ustadmobile.lib.db.entities.Site
 import com.ustadmobile.mui.common.Area
 import com.ustadmobile.mui.common.Sizes
 import dev.icerock.moko.resources.StringResource
 import js.objects.jso
 import mui.icons.material.Chat
```
mikedawson: As per web versin notes: when importing icons always as as NameIcon e.g. as was done for school otherwise IDE performance can be very problematic

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/site/edit/SiteEditViewModel.kt
```diff
@@ -158,6 +175,16 @@ class SiteEditViewModel(
             }
         }
     }
+    fun onTogglePermission(flag: Long) {
```
mikedawson: should be onToggleNavigationItem or something like this - these are not permission flags even though they follow the same pattern

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/login/LoginViewModel.kt
```diff
@@ -110,12 +110,21 @@ class LoginViewModel(
...
-        nextDestination = savedStateHandle[UstadView.ARG_NEXT] ?: ClazzListViewModel.DEST_NAME_HOME
 
         serverUrl = savedStateHandle[UstadView.ARG_LEARNINGSPACE_URL]
             ?: apiUrlConfig.presetLearningSpaceUrl ?: "http://localhost"
         savedStateHandle[UstadView.ARG_LEARNINGSPACE_URL] = serverUrl
 
+        val getDefaultDestinationUseCase: GetDefaultDestinationUseCase =
+            di.on(LearningSpace(serverUrl)).direct.instance()
+        viewModelScope.launch {
+            if (savedStateHandle[UstadView.ARG_NEXT].equals(PersonListViewModel.DEST_NAME_HOME)){
```
mikedawson: I see no reason why PersonListViewModel needs special treatment here- this is a very weird if statement

mikedawson: This is treating the symptom, not the cause.

mikedawson: This is asynchronous code - and it is creating a race condition (automated test might go faster, and next destination would not be correct in that case). This needs to be refactored to avoid a race condition.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/signup/SignupEnterUsernamePasswordViewModel.kt
```diff
@@ -134,6 +137,12 @@ class SignupEnterUsernamePasswordViewModel(
 
                 )
             }
+            if (savedStateHandle[UstadView.ARG_NEXT].equals(PersonListViewModel.DEST_NAME_HOME)){
```
mikedawson: This is copy/pasted duplicated logic (also found in loginviewmodel) - which clearly goes against Do not repeat yourself : https://github.com/UstadMobile/UstadMobile/blob/primary/PULLREQUEST-CHECKLIST.md#very-important (from the very important top part of pull request checklist)

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/signup/SignupEnterUsernamePasswordViewModel.kt
```diff
@@ -206,6 +215,9 @@ class SignupEnterUsernamePasswordViewModel(
             _uiState.value.person?.shallowCopy {
                 this.firstNames = firstName
                 this.lastName = lastName
+                dateOfBirth = savedStateHandle[ARG_DATE_OF_BIRTH]?.toLong()?:0L
```
mikedawson: This belongs in the creation of the entity - it is an initial default being set.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/signup/SignUpViewModel.kt
```diff
@@ -130,6 +135,14 @@ class SignUpViewModel(
         di.on(LearningSpace(serverUrl)).direct.instance()
 
     init {
+        viewModelScope.launch {
+            if (savedStateHandle[UstadView.ARG_NEXT].equals(PersonListViewModel.DEST_NAME_HOME)){
```
mikedawson: As per other comments - the cause should be addressed, we should not copy/paste this if statement to address a symptom into three files.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/site/edit/SiteEditViewModel.kt
```diff
@@ -32,10 +38,13 @@ data class SiteEditUiState(
         listOf(UstadMobileSystemCommon.UiLanguage("en", "English")),
     val fieldsEnabled: Boolean = true,
     val siteNameError: String? = null,
+    val bottomNavToggleError: String? = null,
     val registrationEnabledError: String? = null,
     val currentSiteTermsLang: UstadMobileSystemCommon.UiLanguage = uiLangs.first(),
+    val permissionLabels: List<Pair<StringResource,Long>> = emptyList()
```
mikedawson: These aren't permission labels even though they use the same pattern - call it flagLabels.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/site/edit/SiteEditViewModel.kt
mikedawson: This function should call onEntityChanged; which updates the saved state handle to ensure state is saved if android app destroyed, web user goes back/forth, etc.

core/src/commonMain/kotlin/com/ustadmobile/core/viewmodel/login/LoginViewModel.kt
```diff
@@ -110,12 +110,18 @@ class LoginViewModel(
...
-        nextDestination = savedStateHandle[UstadView.ARG_NEXT] ?: ClazzListViewModel.DEST_NAME_HOME
 
         serverUrl = savedStateHandle[UstadView.ARG_LEARNINGSPACE_URL]
             ?: apiUrlConfig.presetLearningSpaceUrl ?: "http://localhost"
         savedStateHandle[UstadView.ARG_LEARNINGSPACE_URL] = serverUrl
 
+        val getDefaultDestinationUseCase: GetDefaultDestinationUseCase =
+            di.on(LearningSpace(serverUrl)).direct.instance()
+        viewModelScope.launch {
+            nextDestination = savedStateHandle[UstadView.ARG_NEXT] ?:
```
mikedawson, line 123: This is a race condition - no gaurantee that it will be ready first. You could use CompleteableDeferred.

PR #1119, opened 2025-06-18
error message showing now if registration allowed without adding terms

### LINKED ISSUE #1112: Primary Branch || Registration page fails to save without error when Terms and Policies are missing
[](https://devserver3.ustadmobile.com/jenkins/job/primary-sitesaving-error-fix/)

[](https://devserver3.ustadmobile.com/jenkins/job/primary-sitesaving-error-fix-End-To-End-Tests/)

**Describe the bug**
The user is not receiving any error message prompting them to add Terms and Policies before saving.

**To Reproduce**

1. Go to Admin settings.
2. Click on the site
3. Click on the "Edit" button
4. Enable the registration allowed option.
5. Do not add Terms and Policies.
6. Try to save the page.

 **Expected behavior**
An error message should prompt the user to add Terms and Policies before saving.

**Screenshots**

https://github.com/user-attachments/assets/dcecdcd3-3599-437a-8d10-67b5f25c55a6

Nikunjsharma0:


test-end-to-end/webapp-cypress/cypress/e2e/WEB_007_001_admin_enable_or_disable_user_registration.cy.js
```diff
@@ -4,10 +4,36 @@ describe('WEB_007_001_admin_enable_or_disable_user_registration', () => {
...
   // Admin user login
   cy.ustadClearDbAndLogin('admin','testpass',{timeout:8000})
-  cy.ustadEnableUserRegistration()
+  cy.get('#settings_button').click()
+  cy.contains('Site').click()
+  cy.contains('Edit').click()
+  cy.contains('Guest login enabled').should('be.visible')
+  cy.contains('Terms and policies').should('be.visible')
+  cy.contains('Registration allowed').should('be.visible')
+  cy.wait(500) // Without wait switch not clickable
```
poojaustad: I tried adding assertions, a timeout to avoid wait but it's not working so I added wait here

test-end-to-end/webapp-cypress/cypress/e2e/WEB_007_001_admin_enable_or_disable_user_registration.cy.js
```diff
@@ -4,10 +4,36 @@ describe('WEB_007_001_admin_enable_or_disable_user_registration', () => {
...
+  cy.get('#registration_allowed').click({force:true})
```
mikedawson: No workarounds should be needed here. We should not be using force=true. 
If one removes force=true, and then removes wait, component should work as expected.

app-react/src/jsMain/kotlin/com/ustadmobile/view/site/edit/SiteEditScreen.kt
```diff
@@ -84,6 +84,7 @@ val SiteEditComponent2 = FC<SiteEditProps> { props ->
                         }
                     )
                 }
+                error = props.uiState.registrationEnabledError
```
mikedawson, line 89: UstadSwitchField controls - @Nikunjsharma0 needs to add ```disabled = !props.uiState.fieldsEnabled``` (as is already done on text fields 

That will prevent cypress from trying to click it too fast.

mikedawson (approved):

